PR #27, opened 2019-06-20
Fix the references to the old Repository Name

PaulSnow:
Need these fixes to deal with the Repository Rename


Emyrk:
Continued https://github.com/pegnet/PegNet/pull/28

PR #34, opened 2019-06-21
Pnt 17 conversion from fct addresses to peg net asset addresses

PaulSnow:
PNT 17 updates the utilities to convert addresses in a bit easier to manage way.  This pull request adds to the tests.


PaulSnow:
Had to update a reference (that isn't caught by the unit tests I guess).

PaulSnow:
This branch has a few problems.  Taking it back as well.

PR #36, opened 2019-06-22
PNT-35 CLA and license notices

PaulSnow:
Adds an optional CLA, and adds license notices to the files so far


PaulSnow:
Opps.  Got the

PR #38, opened 2019-06-22
Pnt 10 add database

PaulSnow:
Adds a simple database to help calculate balances, past winning records and other stuff for the pegnet.  This is the beginning of an implementation of the wallet.


PaulSnow:
Closing this pull request because it should be against develop
And it assumes too many other pull requests (particularly the license notices, which touch all the files)

PR #41, opened 2019-06-22
Pnt 21 create cli

PaulSnow:
I've implemented the core functionality of the CLI

Of course, we haven't reported balances on the assets yet.  This requires another step, but the CLI is ready to take up that functionality.


PaulSnow:
Closing this pull request because it assumes too many other pull requests that have not yet been reviewed.  Further, it should be against develop, not master.

PR #42, opened 2019-06-23
Pnt 17 conversion from fct addresses to peg net asset addresses fixed

PaulSnow:
This also includes the travis setup


.travis.yml
```diff
@@ -0,0 +1,52 @@
...
+# Get gox cross-compilation library
+# Move into pegnetMining folder for build
+before_install:
+  - go get github.com/mitchellh/gox
+  - cd pegnetMining
+
+
+# Get specific factom branch
+script: 
+  - go get github.com/FactomProject/factom@v1-rollup
```
stackdump: can this be locked using go.mod?

MitchellBerry: quite sure go.mod being included now makes this call redundant
`go get ...@v1-rollup` creates a go.mod if there isn't one already

Emyrk: It should already be locked in the go.mod

Emyrk: I think this was done before go mod was in place

support/uitls.go
```diff
@@ -220,12 +212,39 @@ func ConvertECAddressToUser(addr []byte) string {
 	return base58.Encode(userd)
 }
 
-
 // Convert a User facing Factoid or Entry Credit address
 // or their Private Key representations
 // to the regular form.  Note validation must be done
 // separately!
-func ConvertFctECUserStrToAddress(userFAddr string) string {
+func ConvertUserStrFctEcToAddress(userFAddr string) (string, error) {
```
Emyrk: The factom lib has this, and it checks the checksum too:
https://github.com/FactomProject/factom/blob/master/addresses.go#L43

support/utils_test.go
```diff
@@ -7,19 +7,77 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/FactomProject/factoid"
 	. "github.com/pegnet/pegnet/support"
 	"testing"
 )
 
-func TestBurn(t *testing.T) {
+func TestConvertFctToPegAssets(t *testing.T) {
```
Emyrk: Should probably include hard-coded test vectors to verify addresses do not change in the future.

FCT -> PNT addr list

PR #56, opened 2019-07-01
PNT-51 remove custom sorts

PaulSnow:
This also fixes some bugs in the processing of prior blocks.


opr/grading.go
```diff
@@ -70,52 +71,17 @@ func GradeBlock(list []*OraclePriceRecord) (tobepaid []*OraclePriceRecord, sorte
...
-			if d1 == 0 || d2 == 0 {
-				//panic("Should not be here")
-			}
-			if d1 < d2 { // sort the smallest difficulty to the end of the list
-				list[k], list[k+1] = list[k+1], list[k]
-			}
-		}
-	}
+	// Note that we are sorting in descending order.
+	sort.Slice(list,func(i,j int)bool{return list[i].Difficulty>list[j].Difficulty})
```
WhoSoup: should this be `sort.SliceStable` to guarantee stability? it seems to work out ok in all the situations but there might be some edge cases where it doesn't, or the underlying implementation might change.

PaulSnow: That sort has no need to preserve any prior ordering.  The observation is that difficulties will never match (odds are about 1 in 2^(64-#leading matching bits)).  This would be mitigated by doing the second sort as a standard quicksort, but also mitigated by a stable sort.  The slight performance improvement is likely worth the risk?  But this routine isn't in our inner loop so its performance isn't really interesting.

PaulSnow: I don't think there is any measurable gain to changing the sort, but also no real gain in not changing the sort.  I think it is your call.

opr/grading.go
```diff
@@ -49,79 +52,67 @@ func CalculateGrade(avg [20]float64, opr *OraclePriceRecord) float64 {
...
+			}
+		} else {
+			highest[id] = i
+		}
+	}
+
+	nlist := make([]*OraclePriceRecord, 0)
+	for i, v := range list {
+		id := fmt.Sprintf("%d+%s", len(v.FactomDigitalID), strings.Join(v.FactomDigitalID, "-"))
+		if valid, ok := highest[id]; ok && i == valid {
```
PaulBernier: This condition seems suspicious, is it ever possible that the condition is not true? It looks like if it does, then we have a serious problem in our algorithm and we should crash rather than pass it silently.

Another suggestion: why not iterate through the `highest` map instead to build that `nlist`?

WhoSoup: > This condition seems suspicious, is it ever possible that the condition is not true? It looks like if it does, then we have a serious problem in our algorithm and we should crash rather than pass silently?

it's not possible for `ok` to be false (there's 1+ entry for every id in the list), just a habit of including it in a map check. i suppose it could be removed for readability. `i == valid` is false when there's an entry with the same id but higher difficulty, that's the one to keep

> Another suggestion: why not iterate directly through the highest map instead to build that nlist?
golang maps are iterated through randomly unfortunately. we need to preserve the order

opr/grading.go
PaulBernier: I agree, with both of you. But if I had to really decided, I would have a preference with the stable sort.

WhoSoup: i didn't consider the real life values of the Difficulty field, i was just looking at test data with a lot of potential collisions. i'll probably do some benchmarks to see if there's a noticeable performance difference but otherwise i'd say make it a stablesort

opr/grading.go
PaulBernier: yep got it, thanks. I still need to practice my golang reading, still not completely familiar.

opr/grading.go
```diff
@@ -50,79 +52,64 @@ func CalculateGrade(avg [20]float64, opr *OraclePriceRecord) float64 {
...
-						cont = true // any swap means we continue to loop
-					}
-				}
-			}
-			if !cont { // If we made a pass without any swaps, we are done.
-				break
-			}
-		}
+		// Because this process can scramble the sorted fields, we have to resort with each pass.
+		sort.SliceStable(list[:i], func(i, j int) bool { return list[i].Difficulty > list[j].Difficulty })
```
PaulBernier: Instead of sorting twice it should be doable in one go: https://stackoverflow.com/a/36122804

WhoSoup: i used this in my unit test:
https://github.com/pegnet/pegnet/blob/4b9d0d5f72dcdf5105415560519d7007d564ae29/opr/grading_test.go#L101-L108

PaulBernier: A bit shorter, but a bit less readable (I am big on readability :D). One or the other is fine to me.

Issue #51, opened 2019-06-30
Remove custom sorts


PaulSnow:
We have a few instances where I wrote some quick and dirty sorts.  Replace with calls to sorting libraries.

PaulBernier:
https://github.com/pegnet/pegnet/blob/master/opr/grading.go#L44
I think the whole `GradeBlock` function can be re-written in a way a bit easier to follow, close to the textual description there is in the white paper. But before doing that we would need some test vectors written for the already existing implementation to be able to be sure there is no regression.

I'd be up to rewrite `GradeBlock` function, but preferably someone else should write the tests for the existing implementation (so we will have 2 independent non-biased perspective on this function output).

WhoSoup:
I took a look at creating unit tests for these and it turns out it's a lot of work and there are some bugs in the algorithm, both of them related to duplicate ids.

Bug #1: it picks duplicates
This happens when there are two duplicate IDs and the second one in the list has a higher difficulty. it picks the first one because it's the lowest (the only). it picks the second one because the second one has a higher difficulty than the first.

the algorithm should be altered so that when it finds a new higher record, it removes the previously added record. 


Bug #2: it creates fewer than 10 winners
This happens when there are 10 or more entries in the list but there are enough duplicates that the filtered result contains fewer than 10 entries

you can just move the length check after the duplicate filter

I'm gonna update with my unit test but due to these bugs, I will not cover duplicate identities

PaulBernier:
Great findings @WhoSoup . Even more convinced a clearer rewrite of the grading would be welcomed. If you manager to write a few working tests that would already be a nice win.

PaulBernier:
Ok just noticed that @PaulSnow actually opened a PR for that https://github.com/pegnet/pegnet/pull/56 and it's already much clearer. Paul, when creating a PR you should tag the corresponding issue in its description, it will make it show up in the issue thread and helps visibility. Thanks!

PR #63, opened 2019-07-02
GradeBlock() unit test

WhoSoup:
As requested by @PaulBernier in https://github.com/pegnet/pegnet/issues/51 

Creation of these tests is a bit painstaking but I tried to include as many edge cases as I could. I simplified the input using the same currency value for every single currency, since this is just testing GradeBlock which is about sorting, not Avg() or CalculateGrade().

The tests require at least 10 `gradeEntry` instances per test. `e1` creates an OPR with `difficulty, currencyValue`, and auto-generates miner IDs starting at `"1"`. `e2` adds an additional parameter to set custom miner IDs, to use for duplicate testing further down the road, or non-sequential input.

### DIFF
--- a/opr/grading_test.go
+++ b/opr/grading_test.go
@@ -1,0 +1,456 @@
+// Copyright (c) of parts are held by the various contributors (see the CLA)
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+package opr
+
+import (
+	"fmt"
+	"math"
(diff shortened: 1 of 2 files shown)


PaulSnow:
Man!  Wonderful work!  Painstaking is right!

PaulBernier:
Brilliant, thank you @WhoSoup!

PR #64, opened 2019-07-02
Rework of duplicate check

WhoSoup:
Fixes to the following two issues, as detailed in https://github.com/pegnet/pegnet/issues/51:

> Bug 1: it picks duplicates
> This happens when there are two duplicate IDs and the second one in the list has a higher difficulty. it picks the first one because it's the lowest (the only). it picks the second one because the second one has a higher difficulty than the first.
> 
> the algorithm should be altered so that when it finds a new higher record, it removes the previously added record, or does a double-pass. (ultimately the dupe filter should be moved to its own function)
> 
> Bug 2: it creates fewer than 10 winners
> This happens when there are 10 or more entries in the list but there are enough duplicates that the filtered result contains fewer than 10 entries. this will panic and crash the app

Sidenote: Instead of calculating the actual chain ID for the miners, I'm just using Sprintf to print a unique name. There's a benchmark included in the test file, with these results:
```
BenchmarkSprintVsHash/Sprint_ID-8         	 5000000	       369 ns/op	      74 B/op	       4 allocs/op
BenchmarkSprintVsHash/Sha256_ID-8         	  500000	      2644 ns/op	     506 B/op	      11 allocs/op
```
(test using ids generated from 1-5 words)

### DIFF
--- a/opr/grading.go
+++ b/opr/grading.go
@@ -6,8 +6,11 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"github.com/dustin/go-humanize"
+	"sort"
+	"strings"
 	"sync"
+
+	"github.com/dustin/go-humanize"
 
 	"github.com/FactomProject/factom"
 	"github.com/pegnet/pegnet/common"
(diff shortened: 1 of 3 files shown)


PaulSnow:
I think the approach of using Sprintf is fine, but let's make sure when we actually add Identity registration that we exclude "-" from being used within valid miner identities.  A clean separator is required to uniquely combine a number of fields.  

I want to do this anyway, because we really want to make the IDs 1) human readable, and 2) easy for users to specify without ambiguity.

The Sprintf approach would incorrectly match (fields separated by spaces):

one-two three four => 3+one-two-three-four
one two-three four => 3+one-two-three-four

That's 3 fields but an embedded "-".  Could also be fixed by providing all field lengths, but that requires iteration through the fields and is a bit messy to write.  But it does fix this case:

one-two three four => 3+7+5+4+one-two-three-four
one two-three four => 3+3+9+4+one-two-three-four

[note that the field count isn't necessary if '-' isn't allowed in names, but doesn't hurt anything either.  But maybe it reduces this to just a join?]

WhoSoup:
> The Sprintf approach would incorrectly match (fields separated by spaces):

good catch. if there are prohibited characters in miner names then one of the prohibited characters would suffice.

PR #66, opened 2019-07-03
PNT-44: Added backoff retry strategy when inserting data to factom

gunsmoke:
Also, added a PegExponentialBackOff to common, not sure if it was the best place for it :)

```go
// Default values for PegExponentialBackOff.
const (
	DefaultInitialInterval     = 500 * time.Millisecond
	DefaultRandomizationFactor = 0.5
	DefaultMultiplier          = 1.5
	DefaultMaxInterval         = 60 * time.Second
	DefaultMaxElapsedTime      = 10 * time.Second  // max 10 seconds
)
```

Set the DefaultMaxElapsedTime to 10 seconds for now, this will stop retrying.


opr/oneminer.go
```diff
@@ -52,17 +54,19 @@ func OneMiner(verbose bool, config *config.Config, monitor *common.FactomdMonito
...
 		if err1 == nil && err2 == nil {
-			break
+			return nil
 		}
+		return errors.New("unable to commit Entry to factom")
+	}
+
+	err := backoff.Retry(operation, common.PegExponentialBackOff())
+	if err != nil {
+		// Handle error.
```
PaulBernier: Maybe we can start with a simple logging? With the retry strategy it shouldn't happen much anymore and so shouldn't pollute the log

common/backoff.go
```diff
@@ -0,0 +1,29 @@
...
+func PegExponentialBackOff() *backoff.ExponentialBackOff {
+	b := &backoff.ExponentialBackOff{
+		InitialInterval:     DefaultInitialInterval,
+		RandomizationFactor: DefaultRandomizationFactor,
+		Multiplier:          DefaultMultiplier,
+		MaxInterval:         DefaultMaxInterval,
+		MaxElapsedTime:      DefaultMaxElapsedTime,
+		Clock:               backoff.SystemClock,
+	}
+	b.Reset()
```
PaulBernier: is Reset() really necessary on a just instantiated struct?

common/backoff.go
```diff
@@ -0,0 +1,29 @@
...
+	"time"
+	"github.com/cenkalti/backoff"
+)
+
+// Default values for PegExponentialBackOff.
+const (
+	DefaultInitialInterval     = 500 * time.Millisecond
+	DefaultRandomizationFactor = 0.5
+	DefaultMultiplier          = 1.5
+	DefaultMaxInterval         = 60 * time.Second
```
PaulBernier: a max interval of 60s doesn't make much sense while the max elapsed time is 10s. What about 2s?

common/backoff.go
gunsmoke: That is a good point, I am not sure, kinda followed what was done here 
https://github.com/cenkalti/backoff/blob/v2.1.1/exponential.go#L83

PaulBernier: Ah ok, yeah sure let's keep it then.

gunsmoke:
@PaulBernier Thanks for Review, changes made!

PR #67, opened 2019-07-04
PNT Docker-Compose setup

gunsmoke:
I am used to working with Docker and Docker-compose.

This is my current development setup.

Adding this because it might help others that used docker-compose / docker setup faster.

### DIFF
--- a/Docker.md
+++ b/Docker.md
@@ -9,9 +9,19 @@ To be able to run PegNet you must have Docker and docker-compose, docker version
 - factom-cli installed in your host.
 
 ## Initial Steps
-From wherever you have cloned this repo, run
+From wherever you have cloned this repo
 
-`docker-compose up`
+- Create your config file from `defaultconfig.ini` using
+
+```bash
+cp defaultconfig.ini myconfig.ini
+```
+
+- After you have copied and setup your config file, run
+
+```bash
+docker-compose up
+```
 
 docker-compose will create and run 3 services:
 - factomd (factomd_container)
@@ -23,7 +33,9 @@ You should have factomd and walletd running at this point.
 ## Initial Funding
 Now that we have factomd and walletd running, we need to buy some Entry Credits and fund the default EC address, run
 
-`./Initialization/fundEC.sh`
+```bash
+./Initialization/fundEC.sh
+```
 
 This will fund the EC address *(requires factom-cli)*
 
@@ -31,12 +43,16 @@ This will fund the EC address *(requires factom-cli)*
 
 Make sure the chains are created, run
 
-`docker-compose run --rm pegnet go run Initialization/InitalizeTheNetwork.go`
+```bash
+docker-compose run --rm pegnet go run Initialization/InitalizeTheNetwork.go
+```
 
 # Running the PegNet
 Now that we have everything setup, run
 
-`docker-compose run --rm pegnet go run pegnetMining/Miner.go`
+```bash
+docker-compose run --rm pegnet go run pegnetMining/Miner.go -log debug
+```
 
 # Other usefull commands
 - run `docker-compose up -d` (running in deamon mode)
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY . .
+
+ARG GOOS=linux
+ENV GO111MODULE=on
+
+# Setup the cache directory
+RUN mkdir -p /root/.pegnet/
+COPY defaultconfig.ini /root/.pegnet/defaultconfig.ini
+
+RUN go run Initialization/InitalizeTheNetwork.go
```
PaulBernier: will this work as at this point no EC are available?

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+    depends_on:
+      - factomd
+
+  pegnet:
+    network_mode: host
+    container_name: pegnet_container
+    build: .
+    volumes:
+      - .:/go/src/github.com/pegnet/pegnet
+      - ./defaultconfig.ini:/root/.pegnet/defaultconfig.ini:ro
```
PaulBernier: I am not a Docker expert so I may be not understanding correctly, but in your Dockerfile you already copied that config file, so is it necessary to link this file via a volume?

Docker.md
```diff
@@ -0,0 +1,49 @@
+# PegNet Docker
```
PaulBernier: Could you add a note on this file on how to edit the config file with this set up?

docker-compose.yml
gunsmoke: I do it so I can Edit the default config on the fly ( this allows you to not rebuild the image every time you wanna change that file)

Dockerfile
gunsmoke: This is not 100% necessary and will probably change in the future, this was my way of getting some stuff installed in the image.

docker-compose.yml
PaulBernier: Oh ok, so does it override the file copied in the Dockerfile?

Docker.md
gunsmoke: editing the file should `defaultconfig.ini` should work, we can do it another way

Copy the `defaultconfig.ini` > `myconfig.ini`

and we can change the volume part to
```
volumes:
      - ./myconfig.ini:/root/.pegnet/defaultconfig.ini:ro
``` 

Maybe this is a better solution, what do you thing @PaulBernier ?

docker-compose.yml
gunsmoke: yes, Dockerfile generates the base Image, then docker-compose does magic so we do not need to keep rebuilding the image every change.

Docker.md
PaulBernier: Yep I think it's better. And add a line in the doc explaining that the config file is linked to the host file and that will be neat!

gunsmoke: will do, (will also add this to .gitignore)

gunsmoke:
@PaulBernier Rebased with new changes.

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
+version: '3.2'
+
+services:
+  factomd:
+    network_mode: host
+    container_name: factomd_container
+    image: factominc/factomd:latest
+    entrypoint: /go/bin/factomd
+    command: '--count=5 --net=alot+ --blktime=15 --faulttimeout=20000 --enablenet=false --network=LOCAL --startdelay=2'
```
PaulBernier: Any reason to spawn 5 simulated nodes?
Also I don't remember what does `--net` and `--enablenet` do, what are they and why are they necessary here?

gunsmoke: No particular reason, I was following this script https://github.com/FactomProject/factomd/blob/master/scripts/g25.sh

Please suggest what to change, I am not 100% sure if that is the best settings for local testing =)

PaulBernier: Ok then I would keep it as simple as:
`--network=LOCAL --blktime=15`

gunsmoke: Changed

PR #68, opened 2019-07-04
PNT-60 change grade to look at percent differences

PaulSnow:
Differences in value need to be percent differences to keep large valued tokens like Bitcoin from creating chaos in grading.

Added a feature to slightly randomize the results to make testing easier.


MitchellBerry:
Ok so this build recently failed on windows with the most recent commits. Linux and Mac were fine.
Tried again with the exact same code and it passed. Windows is still in Early Access on Travis and others seem to be running into the same error message. 
I'll modify the script to ignore any windows failures while still testing it.

polling/assets.go
```diff
@@ -67,9 +69,10 @@ type PegItems struct {
 	When  string
 }
 
-func (p *PegItems) Clone() PegItems {
+func (p *PegItems) Clone(randomize float64) PegItems {
 	np := new(PegItems)
-	np.Value = p.Value
+	np.Value = p.Value + p.Value*(randomize/2*rand.Float64()) - p.Value*(randomize/2*rand.Float64())
```
MitchellBerry: Without being seeded first the function `rand.Float64()` will always return the same value for everyone using the default seed of 1

MitchellBerry:
I've run this branch on testnet with no problems.

polling/assets.go
MitchellBerry: something like `rand.Seed(rand.Int63())` at startup is all that's needed

PaulBernier: I am not sure why we have randomization code in the middle production code. If this is for testing, we can find a proper way to have this code extracted and used in the test execution path only. But I don't see a test file in this PR either so I am not sure I understand.

PaulSnow: This is to provide random data for testing the OPR record grading.  We really don't care if it is really random.

PaulSnow: That can be done better.  I'll work on it.

MitchellBerry: `2*rand.Float64` isnt really needed then and can simply be replaced with the hardcoded number `1.2` because that's what it will always return.

Issue #73, opened 2019-07-07
Housekeeping


MitchellBerry:
An issue branch to deal with the small things like linters and gofmt, along with general irregularities that occur along the line. May as well keep it all in the one place.

For example, this:

https://github.com/pegnet/pegnet/blob/ac613a753089add39ac7414b5fe4156834133e37/pegnetMining/Miner.go#L58-L59

gets overridden here: 

https://github.com/pegnet/pegnet/blob/ac613a753089add39ac7414b5fe4156834133e37/common/monitor.go#L142-L143

Believe there's a few lines of unreachable code too somewhere.

MitchellBerry:
Any reason for license headers being after the package? 
I don't get any linting errors from it. Golang source has theirs at the top. Some files are missing them too. 

```go
package pegnetMining

// Copyright (c) of parts are held by the various contributors (see the CLA)
// Licensed under the MIT License. See LICENSE file in the project root for full license information.

import (
```

Emyrk:
License should be above, I'd move it for sure. It's not a linting error, but common practice

Cahl-Dee:
@MitchellBerry so can this be closed? Looks like it was addressed in #97

PR #74, opened 2019-07-08
Lint Housekeeping

WhoSoup:
In the spirit of https://github.com/pegnet/pegnet/issues/73, I'm submitting a bunch of changes to get lint to stop complaining.

Note that https://github.com/pegnet/pegnet/blob/ac613a753089add39ac7414b5fe4156834133e37/opr/opr.go#L37

Was ineffective, and I have updated it to 
```
	FactomDigitalID    []string   `json:"FactomDigitalID"` // [unicode] Digital Identity of the miner
```
which should have no impact on the current json output, since this was supposed to be a non-transformative pull request.

### DIFF
--- a/opr/grading.go
+++ b/opr/grading.go
@@ -10,9 +10,8 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/dustin/go-humanize"
-
 	"github.com/FactomProject/factom"
+	"github.com/dustin/go-humanize"
 	"github.com/pegnet/pegnet/common"
 	log "github.com/sirupsen/logrus"
 	"github.com/zpatrick/go-config"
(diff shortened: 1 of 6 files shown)


opr/grading.go
```diff
@@ -1,22 +1,24 @@
...
 import (
 	"encoding/hex"
 	"encoding/json"
-	"github.com/dustin/go-humanize"
 	"sort"
 	"strings"
 	"sync"
 
+	"github.com/dustin/go-humanize"
+
```
sambarnes: Since its a housekeeping PR anyway, is there a reason for the blank line here? I noticed it in one or two other files changed, where "errors" and "strings" are separated by a blank line. Figure its just your IDE auto adding the imports in a weird place

WhoSoup: yeah, all the imports are auto-handled by my IDE (vscode with go-fmt on save). not sure what exactly caused it. i updated it with a fix :smile_cat:

Issue #65, opened 2019-07-03
Monitor Implementation


WhoSoup:
https://github.com/pegnet/pegnet/blob/4b9d0d5f72dcdf5105415560519d7007d564ae29/common/monitor.go#L13

Two observations:
1. The monitor is never stopped
2. There's only one monitor instance that runs single threaded.

It is possible to just make it a singleton.

There's a lot of superfluous functionality (like stopping) and mutex juggling. I don't think mutexes are even necessary in this case, since the info struct is just swapped out with a new one but there are no goroutines that write data to the fields itself.

Is the stop functionality ever needed in theory? The only instance I can think of is if you hotswap the factomd connection during runtime to a different node that potentially has difference issues with directoryblockheights, but that seems a far fetched scenario. 

Any objections to a slight of this file to reduce mutexes?

MitchellBerry:
Support this. Could certainly do with a trim.
Was originally pulling from the struct to serve up stats but the data is duplicated elsewhere and much more accessible.

WhoSoup:
closing this issue since the pr was merged. thanks everyone!

PR #75, opened 2019-07-08
Monitor Rework

WhoSoup:
As announced in https://github.com/pegnet/pegnet/issues/65, I've reworked the monitor.

There is now a singleton monitor that you can access via `common.GetMonitor()` which initializes and starts the thing. It is possible to start other monitors if you so choose and they won't interfere with each other, but you will have to do it manually.

The giant, confusing loop has been split up into a couple different function calls that should hopefully clear up the functionality and make it easier to test.

There was a bug in the old code that prevented proper error management: 
https://github.com/pegnet/pegnet/blob/ac613a753089add39ac7414b5fe4156834133e37/common/monitor.go#L79-L91

This loop polled factomd node for ~75 seconds, after which it exited with `f.info = nil`. The code immediately afterward accesses `f.info.DirectoryBlockHeight`, which results in a nil-pointer panic. 

I reworked the error management system so the monitor has a timeout, after which it sends the error to the timeout listener. The monitor will continue to keep going and continue throwing errors every specified timeout. At the moment, there is an error listener in `Miner.go` that panics when an error occurs (since it usually means the node is down). If the Timeout is zero, every error is reported.

I also wasn't sure whether to allow rollbacks of heights, which can happen if the underlying factomd node is swapped out at runtime to a node with a lower height. At the moment, it reports all events that are at equal or higher blockheight and don't allow rollbacks. If the endpoint is restarted, the blockheight is the same and the minute rolls back to 0, which is allowed.

I also renamed a bunch of stuff and removed unnecessary info. If there's a need to ever stop and restart the monitor while the node is running, I can re-add that functionality.

### DIFF
--- a/common/monitor.go
+++ b/common/monitor.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/FactomProject/factom"
+	"github.com/cenkalti/backoff"
 )
 
 var monitor *Monitor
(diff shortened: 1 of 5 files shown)


gunsmoke:
@WhoSoup this is looking great.

Backoff was introduced and implemented here 
https://github.com/pegnet/pegnet/pull/66

Do you think we can reuse `backoff` more, in this case 
https://github.com/pegnet/pegnet/blob/ac613a753089add39ac7414b5fe4156834133e37/common/monitor.go#L79-L91

would be a good fit?

WhoSoup:
> Do you think we can reuse `backoff` more, in this case

I added exponential backoff to the bit that queries factomd for the minute, which helped clear up readability for the wait loop. I'm not sure how beneficial the backoff is in this case, however. It only takes effect when the connection to the node breaks, for which it seems Golang already has a 3 second connection timeout. Right now it starts at 3 seconds and then gradually slows down to 10 seconds if it can't find a connection.

common/monitor.go
```diff
@@ -1,158 +1,156 @@
...
+	retry := func() error {
+		info, err = factom.GetCurrentMinute()
+		return err
+	}
+
+	fail := backoff.Retry(retry, bo)
+	if fail != nil {
+		return nil, fail
+	}
+	return info, err
```
PaulBernier: Why do you return that `err` here? If we reached that point it means that we successfully managed to the the current minute. Yet the `err` variable may be not nil if there was at least one failed retry, isn't it?

common/monitor.go
```diff
@@ -1,158 +1,156 @@
...
+}
+
+// MonitorEvent is the data sent to all listeners when being notified
+type MonitorEvent struct {
+	Minute int64
+	Dbht   int32
+}
+
+// Listener spawns a new listening channel that will receive updates of height or minute changes
+func (f *Monitor) Listener() <-chan MonitorEvent {
```
PaulBernier: nitpicking but what about a more explicit `NewListener`?

common/monitor.go
```diff
@@ -1,158 +1,156 @@
...
 
-func (f *FactomdMonitor) GetAlert() chan FDStatus {
-	f.mutex.Lock()
-	defer f.mutex.Unlock()
-	alert := make(chan FDStatus, 10)
-	f.alerts = append(f.alerts, alert)
-	return alert
+// Monitor polls a factomd node and sends alerts whenever the block height or minute changes
+type Monitor struct {
+	polls   int64
```
PaulBernier: I know it's not your code, but I am wondering what conclusion one can draw from knowing how many time we polled? I'd trim that personally.

common/monitor.go
WhoSoup: yeah, it's always nil. updated!

common/monitor.go
WhoSoup: i don't mind nitpicking at all and that is a good suggestion. updated both listener and alertlistener

common/monitor.go
WhoSoup: agreed, removed it.

Issue #21, opened 2019-06-13
CLI Wallet Commands for TXs/Conversions


PaulSnow:
Allow the construction and execution of pegnet conversions and transactions.

sambarnes:
I believe the better way forward is to make pegnet a single binary with multiple options like geth does. This will allow us to do things like:

`pegnet` for normal node/validator running
`pegnet --miners=8` for a node that mines as well

and then while one of the above are running:
`pegnet get balance ADDRESS` can be run to send an RPC command to the local node

This will lower the number of repos we need to have, and also allow users to only have to install a single binary.

Cahl-Dee:
@sambarnes @PaulBernier @PaulSnow Please provide clear requirements as to what will be delivered as part of this issue. It should all be rolled up and placed in the first comment once it has been decided.

Cahl-Dee:
As Sam mentioned in Discord, this is for creating transactions/conversions from the CLI. Transactions and conversions are later milestones and thus this Issue is being pulled from pM1.

Cahl-Dee:
this is addressed in separate issue

PR #76, opened 2019-07-09
CLI Consolidation & General Restructuring

sambarnes:
In response to https://github.com/pegnet/pegnet/issues/21 around creating a separate cli tool to interact with nodes, I think it makes more sense to take the same approach as some other blockchain nodes and have the normal-node/miner-node/wallet all accessible in a single binary.

So I started a reorganization where the entry point is through `pegnet/main.go` and the [cobra package](https://github.com/spf13/cobra) handles the routing of which code paths / components to run.

For example:
-  `pegnet` would run just a validator node (unless overriden in config file to make it a miner
-  `pegnet --miners=8` would run a mining node with 8 miners
- `pegnet createchains` would initialize the various network chains

And we could add `pegnet wallet SUBCOMMAND` for anything we want to add for quick RPC call access

(Aside: I think a better structure might be putting `main.go` in the root of the repo, and move `cmd` folder up a level. And have `pegnetMining` renamed to `miner`, and moving the `opr/oneminer.go` into the `miner` package since its the worker.)

Edit: I see travis failed, I'll update the yaml if people like this new structure


MitchellBerry:
Do like this direction, is there certain miner daemon api endpoints you are looking for?

There's no real reason for main to be in a separate folder, it breaks `go get` just for starters and requires some ide config changes (for me at least)

As for the branch build failing, if anyone is keen the logs here: https://travis-ci.com/pegnet/pegnet/jobs/214467407
Halted before it could begin, seems to be struggling with git?

sambarnes:
@MitchellBerry thanks for checking it out, I agree that main should be moved to the root. Just commited that change.

I did also just update the travis file to just build from root, so that issue is fixed. But I think it broke before that by trying to grab the git branch from pegnet/pegnet instead of sambarnes/pegnet. Unsure though

As for the API endpoints. I'm not sure, likely just balances for now. This ticket is more about just facilitating easy cli access of those endpoints you and rafael are working on

WhoSoup:
i like the idea as a whole

not sure if initialization should be in there since that's something nobody will likely ever run once pegnet is released. it's like a developer-only setting?

sambarnes:
@WhoSoup hmm yeah I think that makes more sense, I'll put it back as its own main package

PR #77, opened 2019-07-10
PNT 22 other polling sources

gunsmoke:
All Sources Call function now returns a Response and have Backoff implemented

~~TODO: implement a simple weight system to choose witch Sources to use~~

### DIFF
--- a/polling/assets.go
+++ b/polling/assets.go
@@ -112,7 +112,7 @@ var prodBlacklistAssets = []string{
 	"OpenExchangeRates",
 }
 
-func GetAssetsByWieght(config *config.Config, assets map[string]func(config *config.Config, peg *PegAssets), default_asset string) []string {
+func GetAssetsByWeight(config *config.Config, assets map[string]func(config *config.Config, peg *PegAssets), default_asset string) []string {
 	var result = []string{}
 	for key := range assets {
 		weight, _ := config.Int("Oracle." + key)
@@ -126,12 +126,12 @@ func GetAssetsByWieght(config *config.Config, assets map[string]func(config *con
 	return result
 }
 
-func GetAvailableAssetsByWieght(config *config.Config) (string, string, string) {
+func GetAvailableAssetsByWeight(config *config.Config) (string, string, string) {
 	rand.Seed(time.Now().Unix())
 
-	var digital_currencies = GetAssetsByWieght(config, availableDigitalAssets, defaultDigitalAsset)
-	var currency_rates = GetAssetsByWieght(config, availableCurrencyAssets, defaultCurrencyAsset)
-	var precious_metals = GetAssetsByWieght(config, availableMetalAssets, defaultMetalAsset)
+	var digital_currencies = GetAssetsByWeight(config, availableDigitalAssets, defaultDigitalAsset)
+	var currency_rates = GetAssetsByWeight(config, availableCurrencyAssets, defaultCurrencyAsset)
+	var precious_metals = GetAssetsByWeight(config, availableMetalAssets, defaultMetalAsset)
 
 	var digital_currencies_asset = digital_currencies[rand.Intn(len(digital_currencies))]
 	var currency_rates_asset = currency_rates[rand.Intn(len(currency_rates))]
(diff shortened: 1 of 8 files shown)


PaulBernier:
OpenExchangeRates gets its data from the European Central Bank, which publishes the rates only once a day. I don't think we should poll that data source given our use case.

gunsmoke:
@PaulBernier Got it, I will remove it then

PS: ExchangeRatesAPI is the same right?

Maybe we can leave `ExchangeRatesAPI` just for Local testing? ( no key required )

PaulBernier:
Oooops. I actually meant `ExchangeRatesAPI` update only once a day and not `OpenExchangeRates`which  can give much more precise data.

We can keep `ExchangeRatesAPI` for testing if you want but we should make sure that in prod it cannot be used. Users won't bother researching the quality of the pre-configured data sources and this could skew the values if they inadvertently pick this one.

MitchellBerry:
At this stage different polling sources seem incredibly useful. Without it the grading mechanism isn't actually being tested, it's all down to hash difficulty. The current testnet exchange rate source requires a paid subscription to contribute to testnet, otherwise the free api limit runs out in a few days, this restricts developer contributions.

As a final outcome you would guess a miner wants complete control over the Oracle values they submit.
For many using general inbuilt sources is likely fine, offering more options doesn't hurt at all.

 In a highly competitive mining environment peg values should be able to be updated at will to anything the miner decides is most likely to win.

polling/exchangeratesapi.go
```diff
@@ -0,0 +1,112 @@
...
+	"net/http"
+)
+
+type ExchangeRatesAPIResponse struct {
+	Date    string
+	Base    string
+	Rates   ExchangeRatesAPIRecord
+}
+
+type ExchangeRatesAPIRecord struct {
```
Emyrk: Can we change this to a map of floats? This just seems klugy.
See what I mean in this playground.
https://play.golang.org/p/Ynmo2ZSC4fj

Then we can write the list of currencies we want, and do it all dynamically.

polling/exchangeratesapi.go
```diff
@@ -0,0 +1,112 @@
...
+	peg.CHF.Value = Round(response.Rates.CHF)
+	peg.CHF.When = timestamp
+	peg.INR.Value = Round(response.Rates.INR)
+	peg.INR.When = timestamp
+	peg.SGD.Value = Round(response.Rates.SGD)
+	peg.SGD.When = timestamp
+	peg.CNY.Value = Round(response.Rates.CNY)
+	peg.CNY.When = timestamp
+	peg.HKD.Value = Round(response.Rates.HKD)
+	peg.HKD.When = timestamp
```
Emyrk: If we do a map of strings above, we can make this a loop over some constant list vs all this repeated stuff.

polling/openexchangerates.go
```diff
@@ -0,0 +1,267 @@
...
+	XDR float64
+	XOF float64
+	XPD float64
+	XPF float64
+	XPT float64
+	YER float64
+	ZAR float64
+	ZMW float64
+	ZWL float64
+}
```
Emyrk: See my comment above about making these things a map

polling/assets.go
```diff
@@ -109,113 +166,44 @@ func PullPEGAssets(config *config.Config) (pa PegAssets) {
...
-		Peg.INR.Value = Round(APILayerResponse.Quotes.USDINR)
-		Peg.INR.When = string(APILayerResponse.Timestamp)
-		Peg.SGD.Value = Round(APILayerResponse.Quotes.USDSGD)
-		Peg.SGD.When = string(APILayerResponse.Timestamp)
-		Peg.CNY.Value = Round(APILayerResponse.Quotes.USDCNY)
-		Peg.CNY.When = string(APILayerResponse.Timestamp)
-		Peg.HKD.Value = Round(APILayerResponse.Quotes.USDHKD)
-		Peg.HKD.When = string(APILayerResponse.Timestamp)
+	
+	digital_currencies, currency_rates, precious_metals := GetAvailableAssetsByWieght(config)
```
Emyrk: Typo, `Wieght` -> `Weight`

gunsmoke: yikes, Nice catch.

PR #78, opened 2019-07-11
Pnt 32 dashboard server

MitchellBerry:
Basic framework for a http api server, still requiring some additional functions and auto--generated documentation, most likely to use go-swagger.


pegnetMining/Miner.go
```diff
@@ -25,6 +27,9 @@ func Mine(numMiners int, config *config.Config, monitor *common.Monitor, grader
 		"miner_count": numMiners,
 	}).Info("Starting to mine")
 
+	http.Handle("/", api.RequestHandler{})
+	go http.ListenAndServe(":8099", nil)
```
sambarnes: I think we should probably move this into the main program entry point over here: https://github.com/pegnet/pegnet/blob/develop/cmd/root.go#L74

So that it runs whenever someone just runs a pegnet node even if they aren't mining. Thoughts?

MitchellBerry: Certainly, makes sense, keeping clutter out of main is a good idea. It could also do with some api versioning too, so a `/v1` route.

MitchellBerry: Added v1 routing and moved server instantiation out of main: c4ac2afea16b209493698232f0b92228feeb3748

sambarnes: Good idea! I'll resolve and merge this in

PR #79, opened 2019-07-11
Pnt 40 ci allow windows failures

MitchellBerry:
Windows builds are still in early access on Travis CI, recently there was a build that only failed on windows, yet I could get the branch to build on a local windows machine. 

This modification will still build for windows and show up in the report but any failure will not affect the final pass/fail grading. Until they stabilise the feature it's probably the easiest way. There's nothing in the codebase that should be affected by platform anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,10 @@ os:
   - osx
   - windows
 
-allow_failures:
-  - os: windows
+matrix:
+  fast_finish: true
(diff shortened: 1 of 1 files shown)


MitchellBerry:
With subtle irony the windows build failed on this PR registering the same error as before.

https://travis-ci.com/pegnet/pegnet/jobs/215277970

Issue #40, opened 2019-06-22
Add a webserver to the pegnetMining to support the CLI


PaulSnow:
Need to be able to handle simple requests as needed by the pn-cli

MitchellBerry:
What type of requests initially? All I can think of:

- toggle miner on/off
- show winning oprs
- get balances

PaulSnow:
Add recording of transactions, but this is likely to be moved over to FAT during a FAT integration effort.

Add recording of conversions, which suggests implementing the processing of conversions as well (though not part of this issue)  This too will likely be moved over to FAT during a FAT integration effort

MitchellBerry:
Docs are up:

https://documenter.getpostman.com/view/8077847/SVSKM98Y?version=latest#69605398-b82a-4cac-be55-bc919a18aed3

sambarnes:
Update these to just have miner statistics (rather than balances)

To be designed in this issue: https://github.com/pegnet/pegnet/issues/117

MitchellBerry:
Would like to simply register that the name of this issue was recently changed.
Is it ok to acknowledge that in the interest of transparency?

MitchellBerry:
The commits made in my name on the pr above were directed at a differently named issue. Think that's more than fair to point out for historical purposes.

sambarnes:
It is logged there already in the chain of events, including what the name was changed from. But yeah explicitly pointing it out is fine too.

I was just told to change the name to something more simpler and more descriptive. Carl didn't like "Add a webserver to the pegnetMining to support the CLI"

Cahl-Dee:
Editted to be about the more fully featured API endpoints that are not required for pM1.

Issue #32, opened 2019-06-21
Control Panel


PaulSnow:
One of the things I really like about factomd is having a control panel that lets us see what's going on with a node.  I am not sure how many crypto daemons for other projects have this feature, but I think it would be great for the PegNet.

If we do keep a form of mining integrated in the PegNet daemon, it would be super cool to be able to turn on and turn off mining right in the daemon.  And it would be great to be able to browse the blocks in the PegNet, and (eventually) the conversons, the transactions, the balances for winning miners, and stuff like that.

Of course, I think the real need will be for a full blown analysis screen that shows charts for the assets, token supplies for assets in the PegNet, and stuff like that.  Is that part of the PegNet Daemon, or something separate?  Likely something separate.

Just going into detail about ideas here.   This Issue is to create the start of a Control Panel with the minimum features:

* Turn on and off mining
* Show the balance of the EC address for writing OPR records
* Show the balances for all Mining rewards by Identity
* Current block height for facotmd and for the pegnet, current minute
* Connectivity to the factomd node supporting the daemon
* Connectivity to the data source used to provide pricing information

PaulSnow:
The Control Panel would be a great place to provide troubleshooting information to help people deploy the pegnet daemon and to set up mining.

PR #85, opened 2019-07-16
PNT-32 Extra endpoints / Cleaning up the api

MitchellBerry:
[API docs online](https://documenter.getpostman.com/view/8077847/SVSKM98Y?version=latest)

Adds:

-  extra endpoints

- cleaner naming system

-  exported functions which may be of use to someone

- reasonably meaningful comments

References:
 - pegnet/pegnet#32
 - pegnet/pegnet#40

Still needed short-term:

- Tests! 

- Raw to PNT address conversions

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -19,17 +19,15 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		jsonDecodingError(w)
 		return 
-  	}
+		}
+	params, _ := json.Marshal(request.Params)
 	log.WithFields(log.Fields{
 		"API Method": request.Method,
-		"Params": request.Params}).Info("API Request")
+		"Params": string(params)}).Info("API Request")
 
 	switch request.Method {
 		case "all-oprs":
-		response(w, Result{OPRBlocks: opr.OPRBlocks})
-		
-		case "balances":
-			response(w, Result{Balances: opr.Balances})
+			response(w, Result{OPRBlocks: opr.OPRBlocks})
 		
 		case "balance":
 			getBalance(w, request.Params)
@@ -37,8 +35,12 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 		case "chainid":
 			response(w, Result{ChainID: opr.OPRChainID})
 
-		case "current-oprs":
-			getCurrentOPRs(w)
+		// case "current-oprs":
+		// 	var current []opr.OraclePriceRecord
+		// 	for _, opr := range opr.CurrentOPRs {
+		// 		current = append(current, *opr)
+		// 	}
+		// 	response(w, Result{OPRs: current})
     
 		case "leaderheight":
 			response(w, Result{LeaderHeight: leaderHeight()})
--- a/api/api.go
+++ b/api/api.go
@@ -5,6 +5,7 @@ package api
 
 import (
 	"bytes"
+	"reflect"
 	"net/http"
 	"encoding/json"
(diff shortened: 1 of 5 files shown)


PaulBernier:
Thank you for putting a high level API doc together, makes it so much easier to review the interface without the details of the code. I will discuss the API purely from the doc as I cannot think of a better place to discuss the interfaces. I haven't look at the code yet.

1. `all-oprs`: "all" with what bounds? All OPRs ever submitted by anyone? Unbounded queries are generally not very usable. Probably need either some filters or some pagination.
2. `chainid`: is that the chain the OPRs are written in? The chain of transactions? Should that method be more explicit in its name, or rather more generic and allow to query different chain ids?
3. `current-rates` returns an OPR, that's rather counter intuitive? Is that different from `winning-opr`? If I understand correctly this actually returns the last winning OPR. So 2 choices: change the API name to directly `winner` (and remove the current winner API that only returns a hash that requires another query) or keep that name but return a shape that actually contains just the latest rates.
4. `leaderheight`: I actually think that all APIs that return "latest X" (`winner`, `winners`...) should be returned with a timestamp (of the block) and the corresponding height. It's not optimal to have to make extra queries to this API to get the height plus it opens up race conditions. With those changes this `leaderheight` API may not be as useful but it doesn't hurt either.
5. I see you started introducting a shorthash. Why did you feel you need thatt? Is see some APIs take shorthash other take full hashes, and you have to make some conversions between them, it's more confusing than beneficial from a user point of view, but maybe I am missing something.
6. ` oprs-by-id ` most likely need to be paginated.

MitchellBerry:
### Pagination:

Certainly, originally had this planned along with a mechanism for the user to restrict concurrency but ended up steamrolling through. Some of the other calls are quite hefty also and will need it too.  

`all-oprs` is essentially the biggest chunk of state stored by the miner right now. Not sure how long it will stay like that underneath, most likely it will move out into the database. It's only going to grow larger regardless. 

How useful the call is in future I'm not sure.

### Chainid:

The one OPRs are written to. 

### Current-rates:

Yep, you are right, they are exactly the same, my mistake, didn't strip out the extraneous data. The name of the call should really line up with what is produced, the naming in general could be better. 

### Leaderheight:

Somewhat redundant call as pegnet needs access to factomd anyway. 

### Shorthash:

Shorthash is simply what I called it there, for interacting with opr's it's a common element and is the only hash natively exported from the structs. A more ideal situation is being able to handle the OPR/Entry/ShortEntry hashes at once and not having to separate them out.

Changing this to v0 or avoiding api versioning altogether for now might be the best idea.

MitchellBerry:
available at `localhost:8099/v0`

Docs are up:

https://documenter.getpostman.com/view/8077847/SVSKM98Y?version=latest

MitchellBerry:
Going to close this while everything plays out, might plug away on it on my own fork. May resubmit depending on how quickly alternatives arise.

Cahl-Dee:
@MitchellBerry is this closed because of the ongoing discussion regarding whether or not we will have the FAT API available?

MitchellBerry:
Yes it was put on hold because of FAT integration. 

The current api isn't in a fully functioning state and this had a few fixes for that but there was a lot of discussion about where exactly the api would be getting served from.

MitchellBerry:
Does FAT have any api code for pegnet I can view or is it more a hypothetical situation at the moment?

I will reopen this PR in one week from now, very keen for any alternatives offered up.

PR #86, opened 2019-07-17
Added burn functionaility to the cli

Emyrk:
Also added the burn address to the common directory. We will need to put the permanent ones there when we decide as to what they will be.
```
pegnet burn FA3EPZYqodgyEGXNMbiZKE5TS2x2J9wF8J9MvPZb52iGR78xMgCb 1
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -112,67 +112,57 @@ var burn = &cobra.Command{
 		// First see if we own the specified FCT address
 		_, err := factom.FetchFactoidAddress(ownedAddress)
 		if err != nil {
-			cmd.PrintErr(err.Error())
-			return
+			CmdError(cmd, err.Error())
 		}
 
 		// Get our balance
 		factoshiBalance, err := factom.GetFactoidBalance(ownedAddress)
 		if err != nil {
-			cmd.PrintErr(err.Error())
-			return
+			CmdError(cmd, err.Error())
 		}
 
 		// Ensure our balance is enough to cover the burn
 		factoshiBurn := factom.FactoidToFactoshi(fctBurnAmount)
 		if factoshiBurn > uint64(factoshiBalance) {
 			fctBal := factom.FactoshiToFactoid(uint64(factoshiBalance))
-			cmd.PrintErrf("You only have %s FCT, you specified to burn %s\n", fctBal, fctBurnAmount)
-			return
+			CmdErrorf(cmd, "You only have %s FCT, you specified to burn %s\n", fctBal, fctBurnAmount)
 		}
 
 		if _, err := factom.NewTransaction(name); err != nil {
-			cmd.PrintErr(err.Error())
-			return
+			CmdError(cmd, err.Error())
 		}
 		if _, err := factom.AddTransactionInput(name, ownedAddress, factoshiBurn); err != nil {
-			cmd.PrintErr(err.Error())
-			return
+			CmdError(cmd, err.Error())
 		}
 		if _, err := factom.AddTransactionECOutput(name, common.PegnetBurnAddress(Network), 0); err != nil {
-			cmd.PrintErr(err.Error())
-			return
+			CmdError(cmd, err.Error())
 		}
 
 		// Signing the tx without a force makes the wallet check the fee amount
 		if _, err := factom.SignTransaction(name, false); err != nil {
 			// Only care about the insufficient fee error here
 			if strings.Contains(err.Error(), "Insufficient Fee") {
-				cmd.PrintErr(err.Error())
-				return
+				CmdError(cmd, err.Error())
 			}
 		}
 
 		// We will force the transaction to ignore any fee too high errors
 		if _, err := factom.SignTransaction(name, true); err != nil {
-			cmd.PrintErr(err.Error())
-			return
+			CmdError(cmd, err.Error())
 		}
 
 		if dryrun, _ := cmd.Flags().GetBool("dryrun"); dryrun {
 			tx, err := factom.ComposeTransaction(name)
 			if err != nil {
-				cmd.PrintErr(err.Error())
-				return
+				CmdError(cmd, err.Error())
 			}
 			fmt.Println("This transaction was not submitted to the network.")
 			fmt.Println(string(tx))
-			return
+			os.Exit(0)
 		}
 		tx, err := factom.SendTransaction(name)
 		if err != nil {
-			cmd.PrintErr(err.Error())
-			return
+			CmdError(cmd, err.Error())
 		}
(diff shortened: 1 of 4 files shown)


cmd/cmd.go
```diff
@@ -92,13 +94,86 @@ var newAddress = &cobra.Command{
...
+	Example: "pegnet burn FA3EPZYqodgyEGXNMbiZKE5TS2x2J9wF8J9MvPZb52iGR78xMgCb 1",
+	// TODO: Verify this functionality.
+	ValidArgs: ValidOwnedFCTAddresses(),
+	Args:      CombineCobraArgs(CustomArgOrderValidationBuilder(true, ArgValidatorFCTAddress, ArgValidatorFCTAmount)),
 	Run: func(cmd *cobra.Command, args []string) {
+		defer factom.DeleteTransaction("burn") // Any cleanup from errors
+		// First see if we own the specified FCT address
+		_, err := factom.FetchFactoidAddress(args[0])
+		if err != nil {
+			fmt.Println(err)
```
PaulBernier: we have a logging tool, let's not use fmt

cmd/cmd.go
```diff
@@ -92,13 +94,86 @@ var newAddress = &cobra.Command{
...
-	Short:"Burns the specied amount of FCT into PNT",
-	Args: CombineCobraArgs(CustomArgOrderValidationBuilder(true, ArgValidatorFCTAddress, ArgValidatorFCTAmount)),
+	Use:   "burn <fct address> <fct amount>",
+	Short: "Burns the specied amount of FCT into PNT",
+	Long: "Burning FCT will turn it into PNT. The PNT burn address is an EC address, and the transaction has " +
+		"an input with # of FCT, and an output of 0 EC. This means the entire tx input becomes the fee. " +
+		"This command costs FCT, so be careful when using it.",
+	Example: "pegnet burn FA3EPZYqodgyEGXNMbiZKE5TS2x2J9wF8J9MvPZb52iGR78xMgCb 1",
+	// TODO: Verify this functionality.
+	ValidArgs: ValidOwnedFCTAddresses(),
```
PaulBernier: What does this do? Are you not exactly checking that in the `Run` function?

cmd/cmd.go
```diff
@@ -92,13 +94,86 @@ var newAddress = &cobra.Command{
...
+	Args:      CombineCobraArgs(CustomArgOrderValidationBuilder(true, ArgValidatorFCTAddress, ArgValidatorFCTAmount)),
 	Run: func(cmd *cobra.Command, args []string) {
+		defer factom.DeleteTransaction("burn") // Any cleanup from errors
```
PaulBernier: "burn" should not be harcoded separately from the variable line 131. Opens up a high risk that one reference is updated without the other

cmd/cmd.go
```diff
@@ -92,13 +94,86 @@ var newAddress = &cobra.Command{
...
+		// First see if we own the specified FCT address
+		_, err := factom.FetchFactoidAddress(args[0])
```
PaulBernier: maybe store args[0] and args[1] is named variables you can reuse, it's easier to read than remember the index of arguments all over this function

cmd/cmd.go
```diff
@@ -92,13 +94,86 @@ var newAddress = &cobra.Command{
...
+			return
+		}
+		tx, err := factom.SendTransaction(name)
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+
+		fmt.Println("Burn transaction sent to the network")
+		fmt.Printf("TransacitonID: %s\n", tx.TxID)
```
PaulBernier: typo

common/common.go
```diff
@@ -12,6 +20,38 @@ const (
...
+		"main": "EC1moooFCT2mainoooo1oooo1oooo1oooo1oooo1oooo1opfDJqF",
+		"test": "EC1moooFCT2TESToooo1oooo1oooo1oooo1oooo1oooo1on1iNDk",
+	}
+)
+
+func PegnetBurnAddress(network string) string {
+	return BurnAddresses[strings.ToLower(network)]
+}
+
+// TODO: Remove this, just making sure w/e burn address we come up with is valid
```
PaulBernier: Usually when reading a "TODO: remove" my immediate question is when? Describe the condition that allows to safely remove that. Otherwise you can bet that no-one will ever take the risk to remove it by fear of breaking something unexpected.

cmd/cmd.go
Emyrk: It's for the auto complete. I'm still trying to to determine the best method to setup autocomplete, as currently you have to run the auto complete installation with walletd up. Then you can tab fill your addresses

Emyrk: It doesn't actually prevent you from using another argument by the way. It just provides a list of valid ones on <tab>

cmd/cmd.go
Emyrk: Do we want to use a logger for a cli error? If you run `pegnet random_command`, the error is printed to stderr. I can move the output to stderr vs stdout, but I'd rather not use logrus

common/common.go
Emyrk: Done :+1:

cmd/cmd.go
Emyrk: Done :+1:

cmd/cmd.go
Emyrk: Done :+1:

cmd/cmd.go
Emyrk: Done :+1:

cmd/cmd.go
Emyrk: I used cobra's print error, which defaults to stderr. I also made the exit code 1 on errors. How do you feel about that?

PaulBernier: logrus outputs everything to stderr (by default) so why couldn't you use logrus for consistency here too?

cmd/cmd.go
PaulBernier: Ok got it, thanks.

cmd/cmd.go
PaulSnow: This is a command line feedback... I really am confused.  I've reviewed, but not merged.  

Paul, can you explain how logrus is a help here?  Doesn't its formatting mess up a command line interface?

Emyrk: I see logrus as a logging mechanism, the cli errors are not logs. They are a cli failure that should notify the user the command line failed to execute and exit with a non zero status.

PaulBernier: Ok ok, I finally got what you meant, my mistake here.

PR #88, opened 2019-07-17
update readme

MitchellBerry:
Some better documentation, the discord widget showing people online will need an admin there to modify a setting. (Clicking on it still takes people to the discord channel)

Markdown rendered here: https://gist.github.com/MitchellBerry/39c88b8fe7ca4db94b928767e331e3d0

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,15 @@ wget https://raw.githubusercontent.com/pegnet/pegnet/master/defaultconfig.ini -P
 
 On first startup there will be some delay while the bytemap is shuffled. It will only start mining at the beginning of a block also.
 
+# Contributing 
+* Join [Discord](https://discord.gg/V6T7mCW) and chat about it with lovely people!
+
+* Run a testnet node
+
+* Create a github issue because they always exist.
+
+* Fork the repo and submit your pull requests, fix things. 
+
 # Development
 
 Docker guide can be found [here](https://github.com/pegnet/pegnet/blob/develop/Docker.md) for an automated solution.
@@ -92,12 +101,3 @@ go build
 
 # You should be ready to roll from here
 ```
-
-# Contributing 
-* Join [Discord](https://discord.gg/V6T7mCW) and chat about it with lovely people!
-
-* Run a node yourself
-
-* Create an issue because it probably exists.
-
-* Fork the repo and submit your pull requests, things always need fixing. 
--- a/README.md
+++ b/README.md
@@ -1,30 +1,103 @@
-# PegNet
-This is now the main repository for mining for the PegNet
+<p align="center">
+  <img src="https://pegnet.org/assets/img/logo.png"/>
+</p>
 
-The TestNet has started writing records in mid June.
+----
 
-This is a work in progress.  The TestNet is certain to reset the actual rewards a few times as we develop more code and adjust the approach.  None the less, it is a great help to have multiple people do mining.  If you would like to help, you can run a miner.  To do so, do these 10 things:
+[![Build Status](https://travis-ci.com/pegnet/pegnet.svg?branch=develop)](https://travis-ci.com/pegnet/pegnet)
+[![Discord](https://img.shields.io/discord/550312670528798755.svg?label=&logo=discord&logoColor=ffffff&color=7389D8&labelColor=6A7EC2)](https://discord.gg/V6T7mCW)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,30 +1,100 @@
...
-1. Clone the OracleRecord repository https://github.com/pegnet/pegnet
 
-2. Copy the defaultconfig.ini file from the OracleRecord repository to ~/.pegnet directory
+## A Network of Pegged Tokens
 
-3. Sign up for an API Key from https://currencylayer.com.
+This is the main repository for the PegNet application.
 
-4. Replace the Oracle.APILayerKey key with your API Key
+Pegged tokens reflect real market assets such as currencies, precious metals, cryptocurrencies, commodities, etc. The conversion rates on PegNet are determined by a decentralized set of miners who submit values based on current market data. These values are then graded and recorded in the Factom blockchain.
```
WhoSoup: Minor rephrase: "These values are recorded in the Factom blockchain and then graded." the other way around could be misleading?

README.md
```diff
@@ -1,30 +1,100 @@
...
+Pegged tokens reflect real market assets such as currencies, precious metals, cryptocurrencies, commodities, etc. The conversion rates on PegNet are determined by a decentralized set of miners who submit values based on current market data. These values are then graded and recorded in the Factom blockchain. 
 
-5. Edit the Miner.IdentityChain to add a unique name like:
+The draft proposal paper is available [here](https://docs.google.com/document/d/1yv1UaOXjJLEYOvPUT_a8RowRqPX_ofBTJuPHmq6mQGQ).
 
-6. IdentityChain=prototype,BigMiner
+For any questions, troubleshooting or further information head to [discord](https://discord.gg/V6T7mCW).
 
-7. Right now you can donate to the cause, and leave the pPNT address alone. We will had a utility to generate your own PNT address from one of your FCT addresses in your wallet
+All mining is currently done on Testnet.
```
WhoSoup: more clarification? it runs on the *PegNet TestNet* which runs on the Factom MainNet

README.md
```diff
@@ -1,30 +1,100 @@
...
+### Manual Environment Setup
+
+Factom binaries
+
+https://github.com/FactomProject/distribution/releases
+
+
+```bash
+# In first terminal
+factomd --count=5 --net=alot+ --blktime=120 --enablenet=false --network=LOCAL # change blktime to whatever suits
```
WhoSoup: why run a simulation with 5 nodes for a sandbox?

README.md
MitchellBerry: True.

Was tempted to go lower level with a description but not sure how general it should be.

README.md
MitchellBerry: Copied from the docker-compose setup I use a custom net in hindsight, perhaps something like that

README.md
```diff
@@ -1,30 +1,100 @@
...
+cd initialization
+go build
+./initialization
+```
+
+### Contributing 
+
+* Run a testnet node
+* Create an issue
+* Write a feature branch on your own fork and submit a PR. Most follow the convention of being named after a particular issue  eg PNT-42_Nifty_Fix
```
WhoSoup: +* Join Discord :smile_cat:

MitchellBerry:
"a network of pegged tokens" ???

¿Does that sound right?

I'm hesitant to use the word stablecoin. Due to certain connotations with others and that it's not necessarily *stable* to a rigidly tight degree compared to the underlying asset. Certainly not in the way something like an ETF is.

README.md
MitchellBerry: A mention of hashing difficulty in grading is probably noteworthy.

README.md
MitchellBerry: By the time this leaves dev branch the statement is probably redundant. A half hearted disclaimer if you will?

README.md
WhoSoup: could also link them to https://docs.factomprotocol.org/start/developer-sandbox-setup-guide in order to avoid redundant explainers, though that might be overkill for some.

maybe a link to the sandbox guide and a "default" command they can use if they don't want to deal with it? something simple like
`factomd -blktime=120 -network=LOCAL`

README.md
WhoSoup: i think the grading process deserves its own separate explainer such as the eventual FATIP protocol description

MitchellBerry:
Updated.

2nd iteration markdown rendering:  https://gist.github.com/MitchellBerry/f58332ccf5246391f65b91ee56d89fc9

Issue #87, opened 2019-07-17
Remove init prints


Emyrk:
Currently running the pegnet cli, we always get prints from init functions in the various packages.

```
2019-07-16 19:23:52.647666143 -0500 CDT m=+0.002079765
Reading ByteMap Table  /home/steven/.lxrhash/lrxhash-seed-fafaececfafaecec-passes-5.size-25.dat

```

MitchellBerry:
This is occurring in the LXRHash repo, and probably the bigger issue is that the hashmap is being initialised for cli calls.

MitchellBerry:
Actually the first timestamp is from here: https://github.com/pegnet/pegnet/blob/28078c9a2a6e7c1dcef50e421522a7ab14e7c960/common/log.go#L21-L24

WhoSoup:
> This is occurring in the LXRHash repo, and probably the bigger issue is that the hashmap is being initialised for cli calls.

related issue: https://github.com/pegnet/LXRHash/issues/25

Emyrk:
https://github.com/pegnet/pegnet/pull/119

PR #90, opened 2019-07-17
Pnt 46 Remove log duplication

MitchellBerry:
Removes duplication of "New OPR Miner" printed for every new miner, now simply one print with the number of miners included.

On startup every previous OPR winner would log. For anyone stopping and starting this fills up files quickly with an obscene amount of info level logs. This fix only prints the winners from most recent block.

Additionally, the very large and small number fields of difficulty/grading are shortened in scientific notation for readability.

Unused and duplicated common.init() function is also removed.

### DIFF
--- a/common/log.go
+++ b/common/log.go
@@ -18,11 +18,6 @@ var DoLogging bool = false
 var start int64
 var LogFile = os.Stdout
 
-func init() {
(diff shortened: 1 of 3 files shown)


MitchellBerry:
partially fixes https://github.com/pegnet/pegnet/issues/87

Issue #101, opened 2019-07-20
Add gofmt to travis 


PaulSnow:
Checking if code is properly formatted (go fmt) is at times very annoying.
We can automate the check.

PaulBernier:
Rather than doing it in Travis after the chance is already pushed, I would recommend to have a git pre-commit hook that run gofmt so that non-formatted cannot be even committed.

Interestingly a quick research yielded this: https://golang.org/misc/git/pre-commit

PR #103, opened 2019-07-21
gofmt checks in travis

MitchellBerry:
also added `go test` and `go test -race` to the CI pipeline 

closes: https://github.com/pegnet/pegnet/issues/101


MitchellBerry:
for some reason a `git update-index --add --chmod=+x` commit wouldn't appear here
 Has no new code just file metadata, probably why, retrying with a new PR

PR #112, opened 2019-07-22
Pnt 102 asset cleanup

Emyrk:
- Move away from the fixed asset list constant structs, and update the asset list to reflect that in the whitepaper.
- Ensure the opr record json is marshaled in the asset order defined (not required, but nice)
- Added unit tests for the apis, using mocked responses for those that require keys.
- Added a default config loader from the code for my unit tests


polling/coincap.go
```diff
@@ -55,52 +58,45 @@ func CallCoinCap(config *config.Config) (CoinCapResponse, error) {
...
+		case "BTC", "XBT":
 			value, err := strconv.ParseFloat(currency.PriceUSD, 64)
-			peg.XBT.Value = Round(value)
-			peg.XBT.When = timestamp
+			peg["XBT"] = PegItem{Value: Round(value), When: timestamp}
 			if err != nil {
 				continue
 			}
-		} else if currency.Symbol == "ETH" {
+		case "BCH", "XBC":
```
MitchellBerry: Coincap api can return different tickers?

MitchellBerry:
Seems like all the miners are cloning `lastanswer` in polling/assets.go after the first round?

None are getting updated data, the `Pulling PEG Asset data` log is never reached after the initial api calls.

common/assets.go
```diff
@@ -0,0 +1,109 @@
...
+ *          Euro,                   EUR,        pEUR
+ *          Japanese Yen,           JPY,        pJPY
+ *          Pound Sterling,         GBP,        pGBP
+ *          Canadian Dollar,        CAD,        pCAD
+ *          Swiss Franc,            CHF,        pCHF
+ *          Indian Rupee,           INR,        pINR
+ *          Singapore Dollar,       SGD,        pSGD
+ *          Chinese Yuan,           CNY,        pCNY
+ *          Hong Kong Dollar,       HKD,        pHKD
+ * DROPPED  Tiawanese Dollar,       TWD,        pTWD
```
PaulBernier: dropped? For what reason? And what's the motivation to keep it in the code then?

common/config.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/go-ini/ini"
+)
+
+//
+type DefaultConfigProvider struct {
+	data string
+}
+
+func NewDefaultConfigProvider() *DefaultConfigProvider {
```
MitchellBerry: Is this meant to replace the config file if not present? Currently it still panics without a defaultconfig.ini

MitchellBerry:
Have run this on a local testnet, all looks good.

common/assets.go
Emyrk: They are listed in the whitepaper, but the data source we poll is missing them. So I removed them, and figured there would need be a conversation as to make a new polling source or actually remove them.

polling/coincap.go
Emyrk: I believe there is a ticker difference from coincap to what we want. But that logic was there before I touched it, I didn't change much actual logic, just making the swap.

common/assets.go
Emyrk: As for motivation, I think it would be good to fix and keep them vs remove them. But that is a business decision that is above me

polling/coincap.go
```diff
@@ -55,52 +57,45 @@ func CallCoinCap(config *config.Config) (CoinCapResponse, error) {
...
 			value, err := strconv.ParseFloat(currency.PriceUSD, 64)
-			peg.LTC.Value = Round(value)
-			peg.LTC.When = timestamp
+			peg["XBC"] = PegItem{Value: Round(value), When: timestamp}
 			if err != nil {
 				continue
 			}
-		} else if currency.Symbol == "XBC" || currency.Symbol == "BCH" {
+		case "ZCASH", "ZEC":
```
Emyrk: Anyone prefer if we use ZEC or ZCASH?

PaulBernier: I think it should be ZEC, the ticker, to remain consistent with the rest.

common/config.go
```diff
@@ -0,0 +1,69 @@
...
+	d := new(DefaultConfigProvider)
+	d.data = `
+[Debug]
+# Randomize adds a random factor +/- the give percent.  3.1 for 3.1%
+  Randomize=0.1
```
WhoSoup: why is this the default?

MitchellBerry:
It's for unit tests.

-------- Original Message --------

common/config.go
MitchellBerry: Unit tests, it won't be created and will still panic if you delete the .ini file

WhoSoup:
had a quick look through and i agree with all the changes. thanks for doing this hard work

common/config.go
Emyrk: @WhoSoup I needed a way to grab the default.ini from the unit tests without needing to deal with file pathing. I grabbed this from the default ini in the repo.

Emyrk: > Unit tests, it won't be created and will still panic if you delete the .ini file

What do you mean by this? Will my unit tests panic under some config conditions?

Emyrk: I renamed it to be more clear

Issue #92, opened 2019-07-18
`Pegnet` Command Miner/API/CLI Organization


Emyrk:
Currently to launch a miner you do `pegnet`, which then reads from the config file to find the number of miners to run. But this is also our cli tool, and soon to be api.

Would you guys be opposed to doing:
`pegnet --service miner --service api` to launch a miner with an api?

then when we add a control panel (if we will?)
`pegnet --service miner --service api --service controlpanel` (or something)
I would also allow the user to define the main-service that takes the main thread, in case for w/e reason they want that.
`pegnet --main-service miner --service api --service controlpanel`
My only concern is that it's not the most intuitive, as I can't think of another app that does this. It's kind of a result of us overloading the `pegnet` binary as being both the cli and the daemon.

I'd rather `pegnet` just display the help of all the available commands, similar to how if you use `docker`.

Does anyone want to defend just launching it as `pegnet` like it is now?

MitchellBerry:
Certainly agree without anything else it should simply print out the help page, fairly standard behavior. Just launching it as is now doesn't feel right, especially with other commands: burn etc. 

The `--service` part seems a bit redundant? (And long to type out) 

Perhaps make the miner a command rather than flag? Guess could get a optional argument in then too. Maybe something like `pegnet mine` to use the config settings or `pegnet mine 12` for 12 miners.

Being able to specify the api or controlpanel seems prudent, people may not want them at all.

MitchellBerry:
Right now the default of no miners gets overridden by the config here: https://github.com/pegnet/pegnet/blob/5abed24b56fb7706a59a54c3cee6e2846b494261/cmd/root.go#L64-L67

`LXRHash.Init()` function should probably be called root.go somewhere too whether or not mining is specified. It shouldn't be getting called unless necessary. A person without the bytemap will have to wait until it's generated first to simply use other commands.

Emyrk:
> Certainly agree without anything else it should simply print out the help page, fairly standard behavior. Just launching it as is now doesn't feel right, especially with other commands: burn etc.
> 
> The `--service` part seems a bit redundant? (And long to type out)
> 
> Perhaps make the miner a command rather than flag? Guess could get a optional argument in then too. Maybe something like `pegnet mine` to use the config settings or `pegnet mine 12` for 12 miners.
> 
> Being able to specify the api or controlpanel seems prudent, people may not want them at all.

Something like that makes sense. I'm not married to how to launch the miner, just to make it not `pegnet`

[...]

PR #113, opened 2019-07-23
Move mining into seperate command

MitchellBerry:
Running the pegnet binary without any arguments will now return the help menu.

To activate the miner:  `pegnet mine`

To use a custom number of miners: `pegnet mine 22`

Using zero or not a number will default to the config setting.

Hashtable won't be initialized now unless mining is specified. 

Tests requiring `LX.Hash()` in future will need to initialize it.

References: https://github.com/pegnet/pegnet/issues/92
Closes: https://github.com/pegnet/pegnet/issues/87

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -5,12 +5,17 @@ package cmd
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/FactomProject/factom"
+	"github.com/pegnet/pegnet/api"
 	"github.com/pegnet/pegnet/common"
+	"github.com/pegnet/pegnet/opr"
+	"github.com/pegnet/pegnet/pegnetMining"
 	"github.com/spf13/cobra"
 )
 
@@ -18,6 +23,7 @@ func init() {
 	// Add commands to the root cmd
 	rootCmd.AddCommand(getEncoding)
 	rootCmd.AddCommand(newAddress)
+	rootCmd.AddCommand(mine)
 
 	burn.Flags().Bool("dryrun", false, "Dryrun creates the TX without actually submitting it to the network.")
 	rootCmd.AddCommand(burn)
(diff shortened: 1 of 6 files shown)


cmd/cmd.go
```diff
@@ -172,3 +178,36 @@ var burn = &cobra.Command{
...
+	Long: "Starts the PegNet mining daemon. Takes number of miners to use as an optional " +
+		"argument, if not specified it will use the default settings from " +
+		"the configuration file",
+	Args:    cobra.MaximumNArgs(1),
+	Example: "pegnet mine 20",
+	Run: func(cmd *cobra.Command, args []string) {
+		Miners = ArgParseMining(args)
+
+		// Initalize hashtable bytemap
+		opr.LX.Init(0xfafaececfafaecec, 25, 256, 5)
```
WhoSoup: we're going with 30 bits, not 25

MitchellBerry: Yeah this should really be a global value.

The grading_tests LX.init should also call the current hash block for correctness also in hindsight.

I won't be able to fix this until tomorrow, if someone can modify the PR am entirely on board.

cmd/cmd.go
```diff
@@ -172,3 +178,36 @@ var burn = &cobra.Command{
...
 	},
 }
+
+var mine = &cobra.Command{
+	Use:   "mine",
+	Short: "Starts the PegNet mining daemon",
+	Long: "Starts the PegNet mining daemon. Takes number of miners to use as an optional " +
+		"argument, if not specified it will use the default settings from " +
+		"the configuration file",
+	Args:    cobra.MaximumNArgs(1),
```
Emyrk: If you make an arg validator like here: https://github.com/pegnet/pegnet/blob/develop/cmd/args.go#L81-L85, you can make sure the argument given is an integer or error out.

For example, this would validate 0 or 1 args, where the first optional arg has to be an integer (assuming you make the validator function)
```
Args: CustomArgOrderValidationBuilder(false, ArgValidatorInteger),
```

cmd/cmd.go
```diff
@@ -172,3 +178,36 @@ var burn = &cobra.Command{
 		fmt.Printf("Transaction: %s\n", tx.TxID)
 	},
 }
+
+var mine = &cobra.Command{
+	Use:   "mine",
```
Emyrk: You have an optional argument for the use, so I'd put something like
```
Use: "mine [miner_count]"
```

MitchellBerry:
new miner rework means this isn't needed now

Issue #110, opened 2019-07-22
Denial of OPR Attack


sambarnes:
Currently, OPRs are not signed by the identity they claim to represent.

Say there are two identities Alice and Bob. And Bob decides to not use his identity in his OPRs and instead list Alice's. However Bob's submissions can effectively block Alice's submissions if he mines with a higher difficulty.

If Bob is honest and submitting good OPR estimates, there is no issue since he'd be helping Alice.

 However, if he chooses to submit terrible estimates as Alice's identity (and mines a higher difficulty than actual Alice was able to do) he can beat out Alice's honest submissions with good estimates. Thus denying her ability to have a competing OPR.

The attacker wins so long as they can prevent another user from being able to compete for rewards.

The above is preventable if we include a signature in one of the ExtIDs that covers all other ExtIDs and the entire json content

MitchellBerry:
Can't see how Alice is prevented from submitting her entries here?

 An ID can submit as many entries as they want to.

WhoSoup:
> An ID can submit as many entries as they want to.

At the moment, OPRs have two fields: the miner ID and address. The grading process throws out entries with a duplicate ID, meaning if you achieve a higher difficulty, you can overshadow someone else's entry with one that has a different address.

The "problem" with signing at the moment is that we eventually want to switch to DIDs once those are working, which may not happen before the pegnet launch, so we'd have to change OPR datastructure fairly soon after launch.

I don't think signing an OPR is necessary though, we just need to change the duplicate filter by incorporating the address as well, so that FA1 + ID1 is not the same as FA2 + ID1. That way we also don't need to generate a new DID for every OPR. The only way someone will be able to submit OPRs in someone else's name will be with the same address, in which case I hope the owner of the address doesn't mind!

see pull request below

MitchellBerry:
I see, why throw out the duplicate ID though? What risk does that pose? Because it's the only differentiating factor in the base LXR Hash?

 The entry hashes for duplicated OPR's have unique entryID's, OPR hash itself is useless after the block is done.

WhoSoup:
> I see, why throw out the duplicate ID though? What risk does that pose? Because it's the only differentiating factor in the base LXR Hash?

The risk is that someone with a very high hashing power can keep all high-POW entries instead of just the highest one, drowning out everyone else.

MitchellBerry:
[...]

PR #114, opened 2019-07-23
Incorporate the address into the duplicate check

WhoSoup:
Pull request for https://github.com/pegnet/pegnet/issues/110

### DIFF
--- a/opr/grading.go
+++ b/opr/grading.go
@@ -6,6 +6,7 @@ package opr
 import (
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


MitchellBerry:
Support this in the checks.

Guess more work still needed to nullify the entire attack vector though, great that it costs a lot more now.

Emyrk:
Travis is failing, can you fix the unit tests that failed?

MitchellBerry:
Failing test:

`grading_test.go:77: RemoveDuplicateMiningIDs() = wrong entry at position 0. got = a2, want = a1`

Was a bit hesitant including go test and race condition testing in the build checks, but think it is for the best.

WhoSoup:
> Failing test:

yeah, the problem with this test is that it's iterating over a map (random) and then trying to sort it back into an expected order, so sometimes i run tests and they're fine and then run them again and they're broken. i'll have to spend some time reworking the test itself rather than trying to apply a quick fix

WhoSoup:
looks like the test finally passed

WhoSoup:
closing on account of not filtering duplicates anymore

PR #115, opened 2019-07-23
Only allow alphanumeric + comma in identities

WhoSoup:
pull request to address https://github.com/pegnet/pegnet/pull/64#issuecomment-507710481
edit: fixed reference

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"os/user"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Emyrk:
Do we want to also enforce the length to be 64 characters? All chains are a 32 byte hash

WhoSoup:
> Do we want to also enforce the length to be 64 characters? All chains are a 32 byte hash

i would say no. they're going to be replaced by DIDs. this is just more or less a sanity check that prevents some confusion 
i just noticed that i tagged the wrong issue in the first post. i meant to tag this discussion: https://github.com/pegnet/pegnet/pull/64#issuecomment-507710481

PR #116, opened 2019-07-23
PNT-81: Remove spaces from OPR chain

sambarnes:


sambarnes:
Yep, I was thinking of making a proper constants file, but figured that'd touch a lot of the code. So probably better as a separate effort

PR #123, opened 2019-07-23
PNT-122 : Made miner it's own struct

Emyrk:
Being it's own struct gives us more flexibility in the code as
we start wanting to track state. The api will begin to tell us
how effective our miners are, meaning having them as trackable
objects is beneficial.

Also:
- Tuned the miner stopping to actually work
- Made `fakes` of the event emitters so we can unit test things that consume the events.


pegnetMining/miner.go
```diff
@@ -0,0 +1,188 @@
...
+				}
+				go oprO.Mine(verbose)
+			}
+		case 9:
+			if mining {
+				close(oprO.StopMining)
+				mining = false
+				writeMiningRecord(oprO)
+				if verbose {
+					did := strings.Join(oprO.FactomDigitalID[:len(oprO.FactomDigitalID)-1], "-")
```
WhoSoup: not a fan of this. i think it's cleaner to just get rid of this at the source: https://github.com/pegnet/pegnet/blob/a39289989dcaadbac5639116e3a759b23f72877c/opr/opr.go#L292-L301

Emyrk: Are you referring to the `did` thing?

WhoSoup: yes. you're splitting FactomDigitalID by "-" and removing the last element ("miner%d") and then replace the last element with your own miner id. I think it'd be cleaner to just remove the code that adds "miner%d" in the OPR creation

Emyrk: I actually pulled this code from the old implementation (not saying it's correct, I'm just still getting up to speed on all the data structures).

That did line is only being used in a log statement. I'm still not sure exactly what you are suggesting.

WhoSoup: > That did line is only being used in a log statement. I'm still not sure exactly what you are suggesting.

oh i missed that part, i assumed it was doing something transformative. there's no change needed in this case, my apologies

PR #129, opened 2019-07-24
PNT-126: Only keep top X miner records

Emyrk:
All miners will collect their top X records, then those lists are merged
and only the top X from the set of miners is kept and written to factom.

Also:
- Made a cmd line flag parser to overwrite configs, so we only need cfg
- Added profiling option


Emyrk:
Ugh, I didn't understand something right. Going to redo this....

PR #130, opened 2019-07-25
Address Parser

WhoSoup:
We'll be needing this at some point and other packages (especially the wallet) will likely find this very useful. The goal is to make a very painless way to handle addresses.

To parse any user input, call `addr, err := common.ParseAddress(input)`, which will take the input and split it into its RCD and prefix. From there, you can convert it into any other pegnet asset representation using: `addr.ToAsset("pXXX")` 

`addr.String()` will return the original string representation that was parsed
`addr.FactomAddress()` will return the factom address (it's the same as `addr.ToAsset("")`)
`addr.Prefix` is the prefix (empty string for FA addresses)
`addr.RCD` are the raw rcd bytes (without checksum)

If you have two different addresses, you can compare them with `addr1.IsSameBase(addr2)` to see if they share the same RCD but with different prefixes.

If you want to change the asset, you can do:
```
addr, err := ParseAddress("FA1y5ZGuHSLmf2TqNf6hVMkPiNGyQpQDTFJvDLRkKQaoPo4bmbgu")
addr.SetPrefix("pPNT")
fmt.Println(addr) // pPNT_111111111111111111111111111111112G1Pfj
```


PaulBernier:
I am not sure I followed all conversations but wasn't `@` preferred over `_` as a separator?

PaulSnow:
I think work on Prefixed addresses (like PNT_xxxxxxx)  should be put on hold at this point.  I don't think creating and using addresses that are understandably tied to assets for miners and users isn't yet on the table.  We are just going to use FAxxxxxxxxxxxxx addresses everywhere in the PegNet, and a set of (separate) tags.  We are being allowed to use tags rather than chain IDs, so there is that.

WhoSoup:
closing on account of not using pnt addresses in core code. maybe one of the eventual wallets can recycle this code

Issue #127, opened 2019-07-24
API Endpoint: "performance"


sambarnes:
In response to https://github.com/pegnet/pegnet/issues/117

Implement the following endpoint:

Method: "performance"

Parameters: 
- `"block_range"`
    - if not present, default to overall performance. Otherwise, must be one of the following:
    - a JSON object `{"start": START_BLOCK, "end": END_BLOCK}` where "end" is optional and defaults to current block (inclusive)
    - if negative values are used for `START_BLOCK` or `END_BLOCK`, it will take the block height that many behind the current directory block head
- `"id"`: the miner ID we wish to request statistics for

Response:
```
{
	"results": {
		"block_range": {
			"start": 201000,
			"end": 2012500,
		},
                "submissions": 1500,
		"rewards": 1337,
                "difficulty_placements": {
                        "1": 5,
			"5": 8,
			"10": 10,
			"25": 20,
			"50": 30,
			"100": 100,
			"250": 200,
			"500": 300,
			"1000": 400,
			"2500": 500,
			"5000": 750,
			"10000": 1000,
			"count": 1500,
			"sum": 125000,
                },
		"grading_placements": {
			"1": 5,
                        "2": 10,
			...
			"49": 28,
			"50": 30,
                        "count": 30,
			"sum": 750,
		},
	}
}
```

Example usage of results:
- `difficulty_placements["count"] / difficulty_placements["sum"` would give you the average rank for your hash power
- `difficulty_placements["10"]  / difficulty_placements["count"]` would give you % of submissions that were in the top 10 hash power

sambarnes:
Branch: https://github.com/pegnet/pegnet/tree/PNT-127_PerformanceAPI/api
Handler: https://github.com/pegnet/pegnet/blob/PNT-127_PerformanceAPI/api/handlers.go#L18

Emyrk:
PR: https://github.com/pegnet/pegnet/pull/133

PR #133, opened 2019-07-25
PNT-127 Performance API

sambarnes:
https://github.com/pegnet/pegnet/issues/127

Unfortunately because of go's json marshaller, the histogram is sorted lexicographically rather than numerically. Maybe there's an easy way to override it? Or a different format?

Example CLI Command:
`pegnet performance prototypeminer001 --start -144 | jq`

Example Request:
```
curl -X POST localhost:8099/v1 --header "Content-Type: application/json" --data '{"method": "performance", "params":{"block_range":{"start":-144}, "miner_id":"prototypeminer001"}}' | jq
```
Example Response:
```JS
{
  "result": {
    "block_range": {
      "start": 7,
      "end": 151
    },
    "submissions": 40,
    "rewards": 18050,
    "difficulty_placements": {
      "1": 7,
      "10": 40,
      "100": 40,
      "1000": 40,
      "10000": 40,
      "25": 40,
      "250": 40,
      "2500": 40,
      "5": 26,
      "50": 40,
      "500": 40,
      "5000": 40,
      "count": 40,
      "sum": 223
    },
    "grading_placements": {
      "1": 4,
      "10": 36,
      "11": 40,
      "12": 40,
      "13": 40,
      "14": 40,
      "15": 40,
      "16": 40,
      "17": 40,
      "18": 40,
      "19": 40,
      "2": 7,
      "20": 40,
      "21": 40,
      "22": 40,
      "23": 40,
      "24": 40,
      "25": 40,
      "26": 40,
      "27": 40,
      "28": 40,
      "29": 40,
      "3": 12,
      "30": 40,
      "31": 40,
      "32": 40,
      "33": 40,
      "34": 40,
      "35": 40,
      "36": 40,
      "37": 40,
      "38": 40,
      "39": 40,
      "4": 15,
      "40": 40,
      "41": 40,
      "42": 40,
      "43": 40,
      "44": 40,
      "45": 40,
      "46": 40,
      "47": 40,
      "48": 40,
      "49": 40,
      "5": 23,
      "50": 40,
      "6": 26,
      "7": 28,
      "8": 32,
      "9": 34,
      "count": 40,
      "sum": 223
    }
  },
  "error": null
}
```

So using these results, you can compute the following stats for prototypeminer001:
- average grading rank when they make it in the top 50: `223 / 40 = 5.575`
- percentage of submissions that made it into the top 5 ranked by difficulty: `26 / 40 = 0.65`


Emyrk:
I think the `difficulty_placements` should only bin to 50, then +Inf. In the future we can optimize to only hash the top 50 honest records, and that would break this api if people rely on it.

Maybe provide some stat like the avg difficulty for the last place? So people can see what the min difficulty is?

sambarnes:
Agree with the reducing of hash power buckets, changing that now. 

Not sure if the average minimum difficulty belongs in this call though, as that is regarding the entire network of miners rather than just the stats of this individual one

PR #148, opened 2019-07-28
Removed the network miner

Emyrk:
It was brought up this was merged in too quickly. This is a pr to back it out


PaulSnow:
Biased, as I did the review, but I don't like removing it at this poin

PR #149, opened 2019-07-28
Do we want the network miner in the main repo?

Emyrk:
Cast your vote if you want this miner in the main repo. 

I did a lot of refactoring around the web services and apis. I was hooking everything up, but if we are not including this in the repo, then I'm not going to spend more time on it.


WhoSoup:
gut feeling says to not include it. it smells of mining pool functionality but it doesn't do as good of a job as dedicated mining pool software would, while opening up potential security issues. (see windows network miners and firewall issues)

i like the idea of someone being able to spin out individual miners to separate machines, but also think that the people who go to that length will switch to dedicated mining pool software when it becomes available

Emyrk:
It's intentionally not a mining pool. It changes the polling behavior to push based. Currently miner's poll exchanges and factomd. This miner gets pushed events, so I made a coordinator to do the polling and push the events

PaulSnow:
I think this is pretty good, because being a CPU mining algo, and having (at least 10 winners) means people don't have to join a pool so much.

If we add some other code to spread out the odds beyond strict difficulty, we can even reach out more to general users.  

What this code does is allow an office or home to install one coordinator on one machine, and collect their best hashes over a large set of miners, and submit them to the network.  If we extend this to mining pools, the coordinator could be the entry point for them as well.

This also solves the issue just running tests on a single computer so we don't pound the factomd node so much.  With a bit of work, this helps to use remote nodes like the Open Node.

Emyrk:
Yea I want to be clear this is out not a mining pool. The current miner is just so poll heavy. I would really want this to be a basic feature option by pegnet. If I move it into it's own repo, then I'd rather just rewrite the miner entirely to be more "mining pool" like.

WhoSoup:
This bit seems like a major security vulnerability: https://github.com/pegnet/pegnet/blob/45134eeb1e5ff46d373c2adbc4551e2ba941e0fd/networkMiner/server.go#L180-L203

Someone could exploit this to spam the factom network using the miner's entry credits (or write any arbitrary data to any chain). The server should at minimum verify that the entry is indeed an OPR before submitting or better yet, send all client submissions to an aggregator to set a global or per-client rate limit.

WhoSoup:
https://github.com/pegnet/pegnet/blob/45134eeb1e5ff46d373c2adbc4551e2ba941e0fd/networkMiner/server.go#L264-L265

it's not! see https://stackoverflow.com/questions/52710882/is-len-thread-safe-in-golang

WhoSoup:
I haven't studied the code intensely but the idea seems nice and the implementation is lightweight enough. If you address the security vulnerability mentioned above and a way to deal with unwanted connections (perhaps a whitelist of ips or a blocklist or a connection password), I think it'd be ok to include this.

Emyrk:
I do plan to make an aggregator per identity. The coordinator should only listen to miners hosted by the same party. This is specifically for a single entity to control both the miner and coordinator. 

I will implement a basic authentication so we reject miners that do not have a shared secret.

Emyrk:
@WhoSoup I added authentication via the form of a shared secret (password) on the miner and the server. They do not send the secret over the network since I'm not using TLS. So they send a hash of a challenge plus the secret. I also fixed the possible race condition on the length of a map.

And some Control panel tweaks.

WhoSoup:
thanks for adding the password challenge! my only nitpick now is that the authentication should be turned on by default :smirk_cat:

Emyrk:
Turned it on by default. The password is `hunter2` by default, so not sure if it buys all that much. But it cannot hurt

PR #151, opened 2019-07-28
Benchmarking GradeBlock

WhoSoup:
These benchmarks take a long time to run (~2 minutes), so here's my findings:

Benchmarking GradeBlock with 10 to 10000 OPR entries:
```
BenchmarkGradeBlock/ten-8                   3000            535030 ns/op
BenchmarkGradeBlock/fifty-8                  100          12260701 ns/op
BenchmarkGradeBlock/200-8                     30          39235576 ns/op
BenchmarkGradeBlock/500-8                     20          91805250 ns/op
BenchmarkGradeBlock/1000-8                    10         176410090 ns/op
BenchmarkGradeBlock/5000-8                     2         878550250 ns/op (~.87 seconds)
BenchmarkGradeBlock/10000-8                    1        1756100400 ns/op (~1.7 seconds)
```

Benchmarking calculating the LXRHash for 10 to 10,000 entries. Please note that for this test, the json is cached before the benchmark is run, so this is purely hashing the available data.
```
BenchmarkOPRHash/opr_hash_for_10-8                   100          21971257 ns/op
BenchmarkOPRHash/opr_hash_for_50-8                    10         103305910 ns/op
BenchmarkOPRHash/opr_hash_for_200-8                    3         411690200 ns/op (~.4 seconds)
BenchmarkOPRHash/opr_hash_for_500-8                    1        1025058600 ns/op (~1 second)
BenchmarkOPRHash/opr_hash_for_1000-8                   1        2067118200 ns/op (~2 seconds)
BenchmarkOPRHash/opr_hash_for_10000-8                  1        21027202700 ns/op (~21 seconds)
```

I picked 10,000 because that's around 16.666 TPS and we had a recent stress test on the testnet with about 15 TPS, so that's a conceivable number we have to deal with.

The cost of the grading algorithm (or rather, the cost of calculating the **difficulty**, which is 90% of the grading work) is negligible compared to the cost of having to calculate the OPR hash in the first place.

### DIFF
--- a/opr/grading_test.go
+++ b/opr/grading_test.go
@@ -5,12 +5,14 @@ package opr
 
 import (
 	"encoding/binary"
+	"encoding/json"
 	"fmt"
 	"math/rand"
 	"sort"
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/FactomProject/factom"
 	"github.com/pegnet/pegnet/common"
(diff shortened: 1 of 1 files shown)


Emyrk:
By self reporting the difficulty, we only need to hash the first 50 honest records, so in a fully honest network, that would keep the number of opr hashes we make down to 50 per block. Which is good. Interesting numbers

PR #153, opened 2019-07-28
Update ByteMap to use 30 bits

WhoSoup:
Not sure why this is still using 25 bits instead of 30.

Alternatively, we could use:
`LX.Init(lxr.Seed, lxr.MapSizeBits, lxr.HashSize, lxr.Passes)`

but i'm not a big fan of that option.


WhoSoup:
There's a problem: Travis will time out generating the 30-bit bytemap. (it takes around 11 minutes on my computer)

WhoSoup:
Alright after a discussion with @MitchellBerry, a solution would be to set the default map size as 30 bits and use an environment variable `LXRBITSIZE` to override the map size to 25 for Travis. 

The downside is that any unit tests that depend on calculating difficulty would have to be done for a specific bitsize. A couple of my unit tests already have lxrhash initialized at specific bit sizes for exactly this reason, but this would be something developers have to watch out for in the future.

PaulSnow:
We can take out the init completely for production.
Put the init of the LXRHash in pegnet.go at root.  Then it is out of the way for Travis.

Unit tests can create LXRHash with 25 bits.

That would avoid the environment variable.

MitchellBerry:
Turns out it's `go test` that's timing out, this can be fixed with a flag. 

We can remove the env variable.

Took 35 mins to run first time around. To speed up the process itself I've added caching of the hashmap file, and took 5 min 2nd time around. So shouldn't be any real overhead from before.

https://travis-ci.com/MitchellBerry/pegnet/builds/121017532

I'll submit it to Who's branch, if you just want to copy/paste the few lines or accept the few dozen inane commits used for testing!

Emyrk:
@MitchellBerry you can cache the entire file? That would be ideal imo

Emyrk:
Looks like the windows build failed there though.

MitchellBerry:
Yeah entire file is cached
All windows builds are failing right now due to gofmt.sh, I'll address that elsewhere.

Emyrk:
@MitchellBerry So this pr is good to merge? It will have the 5m build time?

MitchellBerry:
It still needs the travis.yml diffs from here: https://github.com/WhoSoup/pegnet/pull/1/files

Emyrk:
@MitchellBerry Can we get this Pr against develop? I get it in there, then merge this one in?

MitchellBerry:
I'll put it through now.

Note the first build you do with this pr will take a very long time, it will shorten after that.

-------- Original Message --------

PR #155, opened 2019-07-28
PNT-142 calculate pFCT (and tFCT) balances from FCT burns

PaulSnow:
Run through the factoid blocks, identify the burn transactions, and update the balances appropriately.

Also generalized one of the conversion functions for addressees (we had a FCT to PNT and tPNT conversion, but needed a FCT to pFCT and tFCT)


opr/fct2pFct.go
```diff
@@ -0,0 +1,63 @@
...
+	network, err := c.String("Miner.Network")
+	if err != nil {
+		panic("cannot find the network designation for updating burn txs")
+	}
+
+	if len(OPRBlocks) == 0 {
+		return // There is nothing to do if there is no OPR chain with valid OPR blocks
+	}
+	if FctDbht == 0 {
+		FctDbht = OPRBlocks[0].Dbht
```
WhoSoup: i realize this is an extreme edge case, but the software will `panic` here if the first mined block has no winners or if the chain is empty.

there's a further edge case where fct burns are ignored if they happen during a block without winners before the first block wins

both of these are bootstrap errors that do not occur if the first mined block has winners, but ideally we should start counting fct burns at the "activation height" or the time the time of the first mined block regardless of if it's valid or not

PaulSnow: I debated accepting burns at any height.  Not sure there is a reason to do something else?

PaulSnow: though we know nobody has done a burn for 200k blocks.

PaulSnow: This becomes a question of what happens... If in the end nobody has burned any tokens before we launch the PegNet, then it doesn't matter.

And it depends upon what we think the "launch" of the PegNet is... Is it when the chains are created?  Or when we first mine a block?  I assumed the latter.

WhoSoup: i assume it's an announcement the likes of: "we're going start mining on mainnet at block height 203500" and everyone starts their miners. the edge case is that someone burns FCT in block 203500 but something goes wrong and there are only 5 OPRs in the block. those burnt FCT would be ignored.

i'm not saying we should accept burnst at any height, but we should accept burns the moment that a valid opr block **can exist**

Emyrk: If we start from 0, then we have to parse the entire blockchain to start a pegnet node. I like the activation height as our parsing can skip a lot.

PaulSnow: @WhoSoup Yeah, and I think that is _valid_ but it is just as valid to say burns are valid if they happen after the first valid OPR block on the blockchain.

And if someone burns FCT before the first OPR block, we can always fix this later.  But I'd rather say publicly that nobody should burn assets until we have valid blocks in the Oracle Price Record, and we can even add this check to the burn command in the code.

(Everyone has to use the burn command, because wallets won't let you burn assets beyond a penny or so in value)

WhoSoup: adding a check to the burn command is a good compromise

PR #157, opened 2019-07-29
Moved LXRHash to latest commit

sambarnes:


Emyrk:
People should rename their file to avoid the rebuild

PR #164, opened 2019-07-29
Query coincap by ID vs asset rankings

Emyrk:
Maybe I should move this list out? But it's specific to coincap.io so I put it with coincap.io specifically.

### DIFF
--- a/polling/coincap.go
+++ b/polling/coincap.go
@@ -34,11 +34,37 @@ type CoinCapRecord struct {
 	VWAP24Hr          string `json:"vwap24Hr"`
 }
 
+// CoinCapIOCryptoAssetNames is used by coincapio to query for the crypto we care about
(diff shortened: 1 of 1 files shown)


polling/coincap.go
```diff
@@ -34,11 +34,37 @@ type CoinCapRecord struct {
...
+
 func CallCoinCap(config *config.Config) (CoinCapResponse, error) {
 	var CoinCapResponse CoinCapResponse
 
+	ids := ""
+	sep := ""
+	// Need to append all the ids we care about for the call
+	for _, a := range common.CryptoAssets {
+		ids += sep + CoinCapIOCryptoAssetNames[a]
+		sep = ","
```
PaulBernier: https://golang.org/pkg/strings/#Join

Issue #169, opened 2019-07-30
Grading bugs


PaulBernier:
I believe there are a few important bugs in the grading algorithm that let invalid OPRs be selected. 

The crux of it is in the following `for` loop, that does a couple of odd things: <https://github.com/pegnet/pegnet/blob/develop/opr/grading.go#L212>

- https://github.com/pegnet/pegnet/blob/develop/opr/grading.go#L216 this `continue` statement actually allows (even help!) entries with invalid previous winners to be validated. I believe it should `continue` to the outer loop
- https://github.com/pegnet/pegnet/blob/develop/opr/grading.go#L218 this and the few statements after should not be in this loop, they have nothing to do with `WinPreviousOPR`
- https://github.com/pegnet/pegnet/blob/develop/opr/grading.go#L215 this compares the previous winners of the current OPR with the **previous** winners of the previous oprblock. I believe it should compate it simply with the winners of the previous oprblock.

Emyrk:
Did @WhoSoup Resolve this issue?

PaulBernier:
I believe it did address all the issues, yes

PR #171, opened 2019-07-30
fixing opr validation logic + unit tests

WhoSoup:
this is a fix for the points addressed by @PaulBernier in https://github.com/pegnet/pegnet/issues/169

### DIFF
--- a/opr/grading.go
+++ b/opr/grading.go
@@ -207,27 +207,30 @@ func GetEntryBlocks(config *config.Config) {
 	// the next block we are going to process.
 	// Ignore opr blocks that don't get 10 winners.
 	for i := len(oprblocks) - 1; i >= 0; i-- { // Okay, go through these backwards
-		prevOPRBlock := GetPreviousOPRs(int32(oprblocks[i].Dbht)) // Get the previous OPRBlock
-		var validOPRs []*OraclePriceRecord                        // Collect the valid OPRPriceRecords here
-		for _, opr := range oprblocks[i].OPRs {                   // Go through this block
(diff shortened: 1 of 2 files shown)


PaulSnow:
Wow, very good, and good test.

opr/grading.go
```diff
@@ -207,27 +207,30 @@ func GetEntryBlocks(config *config.Config) {
...
+		for _, opr := range oprblocks[i].OPRs { // Go through this block
+			if !VerifyWinners(opr, previousWinners) {
+				continue
+			}
+			opr.Difficulty = opr.ComputeDifficulty(opr.Nonce)
 
+			f := binary.BigEndian.Uint64(opr.SelfReportedDifficulty)
+			if f != opr.Difficulty {
+				// TODO Maybe we should log.warn how many per block are 'malicious'?
+				log.Errorf("Diff mistmatch. Exp %d, found %d", opr.Difficulty, f)
```
Emyrk: Can we log more context? The ehash and height?

I know I wrote this, but if we are enforcing it, more context would help.
My rational behind the error, was that in the testnet we shouldn't have anyone malicious (unless we mess with ourselves). So if we report this error, there is a bug in our miner/grader that needs addressing. Once we go live, I'm ok with silencing this.

WhoSoup: i thought about it but the rest of the "self reported difficulty" functionality is not actually in yet. i just wanted to limit this PR to fixing the bugs mentioned, and then someone can build on top of it and implement the self-reported feature properly

PaulSnow: That's fair

Emyrk: Ok, that is fine

PR #172, opened 2019-07-30
Do we want to include a script that will cross compile?

Emyrk:
People in the office wanted an windows/mac build. Do we want a script like this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 vendor
 myconfig.ini
+bin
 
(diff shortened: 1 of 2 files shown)


MitchellBerry:
If it works it works. Unsure about the question posed. Artistic codebase minimalism?

gox is alright for cross-compiled builds too:

https://github.com/mitchellh/gox

Emyrk:
Gox looks better :+1:

Issue #170, opened 2019-07-30
Print notification when LXRHash file is being calculated


WhoSoup:
Right now the console is just blank for 10+ minutes with no output while the hash file is being generated. We can either turn LXRHash's verbose option on, or at least print out console output letting people know when it starts and when it ends

Cahl-Dee:
Agree, this is a bit obnoxious.

WhoSoup:
this has been implemented

Emyrk:
Can probably be refined later

PR #178, opened 2019-07-31
lxrhash init print

MitchellBerry:
Prints lxrhash progress on first run.

Removes env variable check for bitsize.

closes: https://github.com/pegnet/pegnet/issues/170


MitchellBerry:
This build is going to fail because of the timeouts still. Looking at the logs the LX.init() is moving incredibly slow.

PR #194, opened 2019-08-01
PNT-193 - Rate Calculation

Emyrk:
The main functions added only calculate the rates if we convert. I did not implement conversions on chain.
 `ExchangeTo` -> Fixed input conversion 
 `ExchangeFrom` -> Fixed output conversion
 `ExchangeRate` -> The rate from asset `FROM` to asset `TO` using USD as the intermediary 

1. Choose 2 random usd rates for `from` and `to`
2. Choose a random amount of `from` tokens (this is `have`)
3. Convert those tokens to `to` for an `amt`
4. Given that amount, how many `from` tokens does it take to get `amt` of `to. (this is `need`).
5. Compare `need` and `have`

`need` and `have` are not always the same. I see differences under 200 sats (most of the time less than 50sats). I provided a set of vectors that have this problem with certain rates and amounts. This means on some conversions, it might be cheaper to choose to fix the input or output by a few sats. I don't know how to solve this, if we switch to int space, we need to choose some large integers to get around this precision. I did not do the math to find how large.

Given the problem is only a few sats, I don't think it is that huge of a concern unless bitcoin or another asset starts getting to the point where sats are very valuable.

### DIFF
--- a/opr/assetList.go
+++ b/opr/assetList.go
@@ -2,6 +2,7 @@ package opr
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 
 	"github.com/pegnet/pegnet/common"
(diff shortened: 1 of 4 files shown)


Emyrk:
Do you guys think we should move to fixed point and big ints? We need to use some pretty large ints to not lose precision and keep the numbers consistent. I'm not opposed to it, and it'd be better to do it now rather than later.

Emyrk:
Actually, this needs some more changes. Will make it an issue

PR #197, opened 2019-08-02
PNT-193 : Rate Calculation

Emyrk:
I began implementing the conversion rates, and want to point out that the implementation has some precision errors. I will describe when these errors arise below.

The main functions implemented added only calculate the rates if we convert. I did not implement conversions on chain.  
`ExchangeTo` -> Fixed input conversion  
`ExchangeFrom` -> Fixed output conversion  
`ExchangeRate` -> The rate from asset FROM to asset TO using USD as the intermediary. I do not round this rate, but the input rates are rounded. Rounding the rate can increase error when the resulting ratio is incredibly small. I don't think we should ever post this rate outside of a conversion, as it will have more significant digits than it's inputs.

1.  Choose 2 random usd rates for from and to
2.  Choose a random amount of from tokens (this is have)
3.  Convert those tokens to to for an amt
4.  Given that amount, how many from tokens does it take to get amt of to. (this is need`).
5.  Compare need and have

The unit test in table format:
```markdown
| TX | From Currency (int64) | Conversion Direction (float64 ratio) | To Currency (int64) |
|----|-----------------------|:------------------------------------:|---------------------|
| 1  | Given `have`          |                 --->                 | Computed `amt`      |
| 2  | Computed `need`       |                 --->                 | Given `amt`          |
| 3  | Computed `whenNeeded` |                 <---                 | Given `amt`         |
```

Results: `need`  and `have` do not match in every case. `whenNeeded` and `need` do match. I believe this is because we lose precision going from `have` to `amt`, and then any math with `amt` will carry that precision error. I think this opens a type of Office Space style thing, where there is sub "sats" being lost/gained on some transactions. Gaming this would be difficult, as you'd have to know the rates, and discover what conversion amts + currencies result in a net gain or net loss of satoshis. The error is limited to <9999 sats I believe, as that is the rounding precision of our current rates. To get that error though, we need an incredibily small resulting rate ratio.
So you might be able to save a few sats by choosing to fix the input or fix the output. An example (you can plug the values into the table). `Difference` is the oppertunity to "gain/lose" sats if making identical txs in opposite directions.
```
# Test Vector
{"FromRate":0.0056,
"ToRate":28.4376,
"ConvertRate":0.000196922384448758,
"Have":49530497071,
"Need":49530499173,
"Difference":-2102,
"WhenNeeded":49530499173}

# Test Vector
{"FromRate":0.0647,
"ToRate":77.2285,
"ConvertRate":0.0008377736198424157,
"Have":62603413597,
"Need":62603413091,
"Difference":506,
"WhenNeeded":62603413091}

{"FromRate":11.7556,
"ToRate":82.6701,
"ConvertRate":0.1421985703254067,
"Have":16158544838,
"Need":16158544840,
"Difference":-2,
"WhenNeeded":16158544840}
```
The error grows as our convert rate gets smaller. The convert rate from usd to btc is ~`0.0001` currently. If anyone has an idea how to increase our precision, or how to solve this inconsistency, I'm all ears.

### DIFF
--- a/opr/assetList.go
+++ b/opr/assetList.go
@@ -3,11 +3,9 @@ package opr
 import (
 	"encoding/json"
 	"fmt"
-	"math/big"
 	"strings"
 
 	"github.com/pegnet/pegnet/common"
-	"github.com/pegnet/pegnet/polling"
 )
 
 // OraclePriceRecordAssetList is used such that the marshaling of the assets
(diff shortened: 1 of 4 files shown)


common/amount.go
```diff
@@ -0,0 +1,79 @@
...
+	if !valid.MatchString(amt) {
+		return 0, fmt.Errorf("improper string")
+	}
+
+	var total int64 = 0
+
+	dot := regexp.MustCompile(`\.`)
+	pieces := dot.Split(amt, 2)
+	whole, _ := strconv.Atoi(pieces[0])
+	total += int64(whole) * 1e8
```
WhoSoup: This regex allows an empty string which produces an error that you're suppressing in line 48. It should be `^[0-9]+(\.[0-9]+)?$`. to force it to be "number", or "number.number". It also doesn't include edge cases like ".5" or "10.". 

Further, your conversion to int ignores all overflow errors which could lead to div by 0 errors or other oddities. It's also not safe for 32-bit systems since you're converting to golang's int.

common/amount.go
```diff
@@ -0,0 +1,79 @@
...
+	return nil
+}
+
+func FloatToAmount(f float64) int64 {
+	return int64(f * 1e8)
+}
+
+// StringToAmount takes a number in decimal form, and converts it to
+// an integer * 1e8. Does not accept negative values
+func StringToAmount(amt string) (int64, error) {
```
WhoSoup: if you don't accept negative values why do you use int64 instead of uint64?

MitchellBerry:
This pr is a bandaid on a bigger problem in my opinion. 

Why using decimal matters:

https://itsfoss.com/a-floating-point-error-that-caused-a-damage-worth-half-a-billion/

https://www.pnsqc.org/dangerously-invisible-failures-floating-point/

https://bitcointalk.org/index.php?topic=13837.0

https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency

https://spin.atomicobject.com/2014/08/14/currency-rounding-errors/

A golang decimal library used in production for half a decade:

https://github.com/shopspring/decimal

 > NOTE: can "only" represent numbers with a maximum of 2^31 digits after the decimal point.

Moving towards an actual release I see no reason why every financial calculation can't use arbitrary precision? It's nothing but trouble otherwise. Not sure if I'd like my name associated with that and formally registering disagreeance.

Why take such risks, this is but one single problem to arise, I'd happily wager there'll be many more. Floats and currencies simply don't mix. The solution is plainly clear as concluded by decades of failures in financial settings. 

oh

common/amount.go
Emyrk: I'll make these changes and add unit tests for those.

common/amount.go
Emyrk: The amounts in the code were all int64. I was just being consistent so I didn't need to do castings if I decided to use this.

Emyrk:
I'd like to point out we are not using floats for balances or regular transactions. Only for conversions. Going from 1 currency to another is going to have lost coins no matter how you roll it. If you roll fixed point, we need to extend our fixed point math for conversions. What I mean by that, is we need to take the balances, accurate to `1e-8`, and increase them to `1e-16` for example, then truncate back at the end of the conversion.

When you are adding or subtracting coins, we still used the fixed point math. Only when we include rates, do we use floats. All of our input floats are also already rounded to the ten thousandth place (I think we should add more resolution there, but that is the current implementation). 

So all inputs are fixed point values where rates are accurate to `1e-4` and balances to `1e-8`. The math to do `balance * rate/rate` is where I enter float space, do the math, then go back to fixed point. If you choose to keep fixed point, you need to increase your precision. That is just required or else large transactions, like `100,000 ADA` which is worth `6,000 USD` would lose 5 decimal places worth of tokens. I don't want to have to redo the math here, I did in on discord, but tl;dr, the `rate/rate` if in fixed point `1e-8`, then by multiplying by `100,000`, you lost all those truncated decimal places that would have contributed to a large ending value. 

If you can provide a concrete reasoning why using floats for the conversion would cause an issue, let me know. But saying "don't use floats because floats are dangerous" is a bit disingenuous. What we need is a deterministic way to convert tokens. Floats are deterministic given the same input. 

I am not:
 - Storing/retrieving floats from a database
 - Using float results into further float calculations beyond the conversion. Addressing the compounding unknown error problem.

What I do agree with:
- Is it possible to overflow an int64 with a balance? I don't know, but we should handle that or investigate that.

Questions:
- Is it easier for other language implementations to implement a float based conversion or a decimal?
- Can you provide a concrete example that using a float will hurt us? Can you then provide decimal result? (I did this on discord with the 1e-8 precision on the decimal, and lost more tokens in my conversion by a decent amount)

common/amount.go
Emyrk: Wait, I have that regex above. This is just for the splitting, if there is no decimal point, then the split results in 1 sections.

`10.` is deemed invalid though.

I added better overflow checking

opr/assetList.go
```diff
@@ -17,9 +23,80 @@ func (o OraclePriceRecordAssetList) Contains(list []string) bool {
...
 }
 
+// Exchange tells us how much we need to spend given the amount we want is fixed.
+//	?? FROM -> X TO
+//
+//   X TO         to_usd               1
+//  ------    *  --------- = USD * -------- = FROM
+//     1            1               from_usd
+//
+func (o OraclePriceRecordAssetList) ExchangeTo(from string, to string, want int64) (int64, error) {
```
WhoSoup: both this and exchangefrom seem to only operate on whole numbers, which seems kinda useless. is this supposed to only work for the smallest denomination, aka cents / sats / whatever? if so, it really needs to be pointed out

also the names for ExchangeTo and ExchangeFrom are really confusing. took me a long time to figure out what they're supposed to do. they should be swapped around

i'm not sure we even need the function for "how much X do i need to get Y?", since that's literally just swapping the order of the arguments

opr/assetList.go
```diff
@@ -17,9 +23,80 @@ func (o OraclePriceRecordAssetList) Contains(list []string) bool {
...
+
+	// fromRate / toRate
+	return fromRate, toRate, nil
+}
+
+// ExchangeRate finds the exchange rate going from `FROM` to `TO`.
+//	To do the exchange rate, USD is the base pair and used as the intermediary.
+//	So to go from FCT -> BTC, the math goes:
+//		FCT -> USD -> BTC
+func (o OraclePriceRecordAssetList) ExchangeRate(from, to string) (float64, error) {
```
WhoSoup: this function is not compatible with the above ExchangeTo/From. the above functions multiply first, then divide, whereas this function divides first, meaning you get different results

assuming we have:
1 FTC = 8 USD
1 BTC = 123456.12345678 USD

and we want to convert 6,838,893.09091930 FTC to BTC. 

```go
o := make(opr.OraclePriceRecordAssetList)
o["USD"] = 1
o["FCT"] = 8
o["BTC"] = 123456.12345678

fmt.Println(o.ExchangeFrom("FCT", 683889309091930, "BTC"))
rate, _ := o.ExchangeRate("FCT", "BTC")
fmt.Println(common.TruncateFloat(float64(683889309091930) * rate))
```
the result is:
```
44316266537
44316266536
```
meaning a 1 sat difference. we really need to use a dedicated library for this kind of math

Issue #198, opened 2019-08-03
Lack of feedback when launching miner


PaulBernier:
when launching `./pegnet` here is the output at the begining:
```
Reading ByteMap Table /home/paul/.lxrhash/lxrhash-seed-fafaececfafaecec-passes-5-size-30.dat
INFO[0000] Starting control panel on localhost:8080     
Finished Reading ByteMap Table. Total time taken: 1.196095765s
```
and then nothing for few minutes (until a new block I believe?). This lack of feedback is confusing for the end user who has no idea what is going on or even if the miner is working or not.

Emyrk:
@PaulBernier What are you thinking to include? Would something like this be sufficient?
```
Reading ByteMap Table /home/steven/.lxrhash/lxrhash-seed-fafaececfafaecec-passes-5-size-30.dat
INFO[0000] Starting control panel on localhost:8080     
Finished Reading ByteMap Table. Total time taken: 442.778981ms
INFO[0000] Miners launched. Waiting for minute 1 to start mining...  id=coordinator
INFO[0008] New OPR Winner                                address=tPNT_yFQZ8o1d8n5PKJAm43qNy4b6ynYwiFEq7R2pVgkuHy41qcozx balance="30,580,000,000,000" difficulty=1.8446692497e+19 entry_hash=dcbc2be9546699df grade=2.1203e-62 id=factomEmyrk place=0
INFO[0008] Pulling PEG Asset data                        delta_time=1565016700
INFO[0011] Begin mining new OPR                          height=75 id=coordinator minute=1
```

I added
```
INFO[0000] Miners launched. Waiting for minute 1 to start mining...  id=coordinator
```

And this on every minute 1:
```
INFO[0011] Begin mining new OPR                          height=75 id=coordinator minute=1
```

PaulBernier:
Merged, thank you.

PR #201, opened 2019-08-05
PNT-198 : Added more miner feedback when starting

Emyrk:
From https://github.com/pegnet/pegnet/issues/198

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -267,3 +267,10 @@ func ConvertFCTtoPegNetAsset(network string, asset string, userFAdr string) (Peg
 	}
 	return
 }
+
+func Abs(v int) int {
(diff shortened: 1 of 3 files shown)


common/utils.go
```diff
@@ -267,3 +267,10 @@ func ConvertFCTtoPegNetAsset(network string, asset string, userFAdr string) (Peg
 	}
 	return
 }
+
+func Abs(v int) int {
```
PaulBernier: :smile: http://cavaliercoder.com/blog/optimized-abs-for-int64-in-go.html golang programmers...

Emyrk: Wow. I need to optimize it lol

PR #208, opened 2019-08-08
PNT-205 : polling stablility

Emyrk:
Currently I have it setup such that you choose the priority of your data sources, and I fallback the less priority ones if the first choice fails. So if you request the order is:

```
SourceA = 1
SourceB = 2
SourceC = 3 
```

Then I fetch all the assets running through those in order. If an asset fails, I go to the next. A -> B -> C. I only make max 1 api call per exchange, so the exchange only gets called once if I end up needing it. I made the precious metals from OpenExchangeRates work, and that data source supports BTC. So if you List OpenExchangeRates above coincap, you will get BTC from open exchange, and the rest of crypto from coincap/coinmarketcap. If the metals from OpenExchange fails, then we try Kitco.

You do do specific overrides too if for example, you want to get BTC from the crypto sources over Open exchange rates, even if open exchange has a higher priority.
```
[OracleAssetDataSourcesPriority]
  # Specific coin overrides
  XBT=CoinMarketCap,CoinCap,OpenExchangeRates
```


Emyrk:
@MitchellBerry Yea, I changed the polling entirely. I did keep the existing code that polls each exchange though. I didn't feel like changing those, and they worked. Just changed where they fed into and added CMC

MitchellBerry:
it's really good, there's some great changes in there
was just a general call for eyeballs on it

Issue #232, opened 2019-08-13
Add Factomd and Walletd host:port to config file options


afenrir:
**Request**
Currently, it is necessary to pass the factomd and walletd host and port via CLI args if they are not in their default locations. It would be useful if those could be set in the config.ini file instead of/as well as passed via CLI args. 

**Rationale**
It will not be immediately obvious to the end user what the correct host and port options are for walletd and factomd if the miner is brought up with docker-compose, particularly if the user has little or no experience of Docker. 

Setting default options for the factomd/wallet host/port in the config file on the pegnet-compose repo will allow me to hide those implementation details from the user. Otherwise, each time the user overrides the default run command (for example, to change the number of miners), they will need to pass the correct networking options at the same time.

PR #233, opened 2019-08-14
adding factom/wallet location to config

WhoSoup:
close: #232 

adds Miner.FactomdLocation and Miner.WalletdLocation. they get overwritten by command line.


WhoSoup:
note: this also fixes nil-pointer bug where it tries to ++ before checking if there's an error 
https://github.com/pegnet/pegnet/blob/966a70475d10dc95f50b76b4a4c2907fbb8070e7/common/monitor.go#L102-L107

Issue #237, opened 2019-08-15
Grader should save some state to a database


Emyrk:
The current grader sync time is quite slow, we should be able to save some state to speed this up

Issue #236, opened 2019-08-15
Grader only grade top 50 honest oprs


Emyrk:
Optimization to reduce some opr hashing

PR #238, opened 2019-08-15
Grader refactor

Emyrk:
The grader has been redone. The previous iteration would regrade some blocks and it used a global as the oprblock storage. This refactor moves the global state into a struct and will only grade each entry block once. 

## Grader Syncing
The basics of the new grader is that we first sync the eblocks from the network. Once we have the eblocks in their sequence, we can sync each eblock one by one, marking them synced as we walk through the chain. This style of syncing was used in factomd for identities, as you only sync heights with data we are concerned about (since we hope eblock to eblock, ignoring the dblocks). It makes keeping track of a sync simpler, as the prior code would use a sorted block height order to verify the order of oprs was correct. This uses the natural chaining of eblocks, and handles height skips gracefully.

This eblock syncing behavior was moved into it's own struct if we wish to reuse it for another chain syncing.

## Grader Optimiziation

The grading used to compute the difficulty of all the oprs in a block, now we only do the top 50 honest records. With the change to sha256, this is less important, but can really help speed up syncing as the chain grows.

## Grader Database

I didn't want to save the entire state to the database, as that could be prone to more errors, and did not want to undertake that at this time. However, grader syncing is already starting to take a good amount of time if using a remote factomd node (15m on the older grader to a factomd over LAN). I added multithread entry retrival and got that sync time down to 3min, but 3min for each boot is still slow. There is now a database that stores the graded oprs for a given eblock, making the sync time 6s to my factomd across LAN after it's been synced once. This can be further speedup by saving the eblock syncing part, but I did not feel that necessary atm.

## Further notes

I think we should drop a lot of the state we are saving in the miner, mainly to support the apis. I suggest we create a different daemon `pegnet node` that can store all the information to support the apis. We are overloading the miner, and increasing the memory footprint for a miner that might be running in low resource environments. If we drop the api, and instead support a very basic set of endpoints, we can create a better pegnet node with better database indexing for the rest of people's needs.

This is the start of splitting the `miner` and the `node`. I want to get away from the globals so we can easily split things apart and not be sharing globals where the data format is fixed.

Addresses issues:
- https://github.com/pegnet/pegnet/issues/236
- https://github.com/pegnet/pegnet/issues/237

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -4,8 +4,6 @@
 package api
 
 import (
-	"bytes"
-	"encoding/hex"
 	"strconv"
 
 	"github.com/FactomProject/factom"
@@ -15,7 +13,7 @@ import (
 // -------------------------------------------------------------
 // Required for M1
 
-func getPerformance(params interface{}) (*PerformanceResult, *Error) {
+func (a *APIServer) getPerformance(params interface{}) (*PerformanceResult, *Error) {
 	performanceParams := new(PerformanceParameters)
 	err := MapToObject(params, performanceParams)
 	if err != nil {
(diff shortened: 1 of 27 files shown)


Emyrk:
Rebasing -> `grader_refactor_rebase`

PR #247, opened 2019-08-17
change default coinbase to something invalid, added checking in code,…

PaulSnow:
Fix some config.ini file issues that slipped through the cracks.

The really big one is that we can't default the coinbase address to something, or miners will lose money.

Also added quick error checking so miners get immediate feedback.

Added comments

Added a check for config.ini before defaultconfig.ini because the latter as a config file name is just stupid.


PaulBernier:
The master branch already contains the change `CoinbaseAddress=CHANGEME`. Closing this PR.

PR #251, opened 2019-08-18
Price inversion was happening twice, nullifying it

Emyrk:
Made the unit test for price checking also on the fixed price unit tests


polling/openexchangerates.go
```diff
@@ -112,7 +112,7 @@ func CallOpenExchangeRates(c *config.Config) (response OpenExchangeRatesResponse
 	// Price is inverted
 	if err == nil {
 		for k, v := range OpenExchangeRatesResponse.Rates {
-			OpenExchangeRatesResponse.Rates[k] = 1 / v
+			OpenExchangeRatesResponse.Rates[k] = v
```
PaulBernier: the whole loop can actually be removed now

Emyrk: Removed it, as long as the api layer one too

PR #260, opened 2019-08-21
Fixed `docker-compose` file to work with Windows

dwjorgeb:
Added `factomd.conf` file to point to the right daemon containers
Moved `defaultconfig.ini` file to `config/` dir
Updated documentation to reflect the changes
Fixed `Dockerfile` file to point to new config path

### DIFF
--- a/Docker.md
+++ b/Docker.md
@@ -11,10 +11,16 @@ To be able to run PegNet you must have Docker and docker-compose, docker version
 ## Initial Steps
 From wherever you have cloned this repo
 
-- Create your config file from `defaultconfig.ini` using
+- Create your config file from `config/defaultconfig.ini` using
 
 ```bash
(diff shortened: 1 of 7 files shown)


Docker.md
```diff
@@ -61,9 +67,9 @@ Or a node with a set number of miners:
 docker-compose run --rm pegnet go run main.go --log=debug --miners=4
```
afenrir: `main.go` is now `pegnet.go` on develop. Please would you change this?

dwjorgeb: done :)

afenrir: Sorry, my bad, line 60 also references main.go. Would you mind changing that too? Other than that problem, I have tested this and it works, so will merge as soon as that is changed.

dwjorgeb: whoops, missed that one. Fixed @afenrir

PR #274, opened 2019-08-26
PNT-239 : `pegnet node` 

Emyrk:
Has a few things, the primary goal is to separate the `node` from the `miner`. The miner has not been touched as far as features, this is mainly adding the `node`. This is currently powering factoshi.io.

# Factoshi.io stats & new api

`pegnet node` is the new source of the apis. It has a SQLite database, which lives in a subdirectory of `~/.pegnet/data_MainNet/`. The `MainNet` is so if you run `TestNet` a different filepath is used, and you do not need to delete/move/configure anything. The node extends the existing api endpoints, so it's the same port as before, just some additional methods on a new url endpoint `/node/v1`. 

## API Changes
 - Performance api now also works for payout addresses too, not just identities

# Grader Refactor
The grader has been redone. The previous iteration would regrade some blocks and it used a global as the oprblock storage. This refactor moves the global state into a struct and will only grade each entry block once. 

## Grading Algorithm change

Activation height with grading algorithm change.
- XPD and XPT dropped
- More precision in reported prices

### Grading Algo
1. Pow to top 50
2. Grade with 0.5% band to top 25
3. Pay top 25 200PNT each
4. Grade to 1
5. Final winner is the price

## Grader Syncing
The basics of the new grader is that we first sync the eblocks from the network. Once we have the eblocks in their sequence, we can sync each eblock one by one, marking them synced as we walk through the chain. This style of syncing was used in factomd for identities, as you only sync heights with data we are concerned about (since we hope eblock to eblock, ignoring the dblocks). It makes keeping track of a sync simpler, as the prior code would use a sorted block height order to verify the order of oprs was correct. This uses the natural chaining of eblocks, and handles height skips gracefully.

This eblock syncing behavior was moved into it's own struct if we wish to reuse it for another chain syncing.

## Grader Optimiziation

The grading used to compute the difficulty of all the oprs in a block, now we only do the top 50 honest records. With the change to sha256, this is less important, but can really help speed up syncing as the chain grows.

## Grader Database

I didn't want to save the entire state to the database, as that could be prone to more errors, and did not want to undertake that at this time. However, grader syncing is already starting to take a good amount of time if using a remote factomd node (15m on the older grader to a factomd over LAN). I added multithread entry retrival and got that sync time down to 3min, but 3min for each boot is still slow. There is now a database that stores the graded oprs for a given eblock, making the sync time 6s to my factomd across LAN after it's been synced once. This can be further speedup by saving the eblock syncing part, but I did not feel that necessary atm.

## Further notes

I think we should drop a lot of the state we are saving in the miner, mainly to support the apis. I suggest we create a different daemon `pegnet node` that can store all the information to support the apis. We are overloading the miner, and increasing the memory footprint for a miner that might be running in low resource environments. If we drop the api, and instead support a very basic set of endpoints, we can create a better pegnet node with better database indexing for the rest of people's needs.

This is the start of splitting the `miner` and the `node`. I want to get away from the globals so we can easily split things apart and not be sharing globals where the data format is fixed.

Addresses issues:
- https://github.com/pegnet/pegnet/issues/236
- https://github.com/pegnet/pegnet/issues/237


opr/grader.go
```diff
@@ -104,23 +183,416 @@ func (g *Grader) Run(config *config.Config, monitor *common.Monitor) {
...
+	}
+
+	var oprs []*OraclePriceRecord
+	var err error
+	if len(block.EntryBlock.EntryList) > 50 {
+		// Multithread when there is a lot (like a 6x speedup in my tests against mainnet)
+		oprs, err = g.ParallelFetchOPRsFromEBlock(block)
+	} else {
+		oprs, err = g.FetchOPRsFromEBlock(block)
+	}
```
WhoSoup: why have two separate implementations for the same code? just have one function that specifies the number of workers and set it to 1 if you want single threaded. is there a reason to not just run the workers every time?

opr/grading.go
```diff
@@ -46,18 +48,145 @@ func Avg(list []*OraclePriceRecord) (avg []float64) {
...
+func gradeMinimumVersionTwo(sortedList []*OraclePriceRecord) (graded []*OraclePriceRecord) {
+	list := RemoveDuplicateSubmissions(sortedList)
+	if len(list) < 25 {
+		return nil
+	}
+
+	// Find the top 50 with the correct difficulties
+	// 1. top50 is the top 50 PoW
+	top50 := make([]*OraclePriceRecord, 0)
+	for i, opr := range sortedList {
```
WhoSoup: this needs to iterate over `list` (post duplicate filter) not `sortedList` (pre duplicate filter)

opr/grading.go
```diff
@@ -46,18 +48,145 @@ func Avg(list []*OraclePriceRecord) (avg []float64) {
 }
 
 // CalculateGrade takes the averages and grades the individual OPRs
-func CalculateGrade(avg []float64, opr *OraclePriceRecord) float64 {
+func CalculateGrade(avg []float64, opr *OraclePriceRecord, band float64, version uint) float64 {
```
WhoSoup: why is `version` a parameter if it's not used?

opr/recorder.go
```diff
@@ -28,9 +35,7 @@ func NewChainRecorder(con *config.Config, filpath string) (*ChainRecorder, error
 	return c, nil
 }
 
-// WriteMinerCSV will write all the miner related stats for a given chain to a csv. This includes
-// number of records, difficulties hit, and more.
-func (c *ChainRecorder) WriteMinerCSV() error {
+func (c *ChainRecorder) WritePriceCSV(db database.IDatabase, height int64) error {
```
WhoSoup: initlx() can be removed from this function since it's not specifically used by that function. quickgrader inits it

WhoSoup:
> The `MainNet` is so if you run `TestNet` a different filepath is used, and you do not need to delete/move/configure anything

how does this handle local network vs testnet, or a local mainnet vs mainnet mainnet? any way to get network id in there too?

WhoSoup:
re: band of 1%

this is from my early analysis of 8 apis (alternativeme, freeforexapi, openexchangerates, apilayer, fixedusd, coincap, coinmarketcap, 1forge) tracked over 4.5 hours. i calculated the maximum difference between two sources as percentage to see how they match up with each other. anything that crosses over the 0.01 line would not fit inside a 1% band and miners using that value would likely be graded out

image

all the lines crossing 1% are crypto (ZEC, XMR, RVN, etc) with the main culprit being a difference in alternativeme vs coinmarketcap, and alternativeme vs coincap. unfortunately i had a bug in my code that didn't save finer granularities, so i can't tell you yet how it would watch up without alternativeme. i'll have the full data by tomorrow

here's the raw data (values over 0.01 are highlighted, values over 0.04 in red) with a clickable chart so you can actually tell which line is which
https://docs.google.com/spreadsheets/d/13PSNcKnCw_99IzXVQgeQ9yGM8HPdFkNcczLPWhmiMUk/edit?usp=sharing

opr/opr.go
```diff
@@ -139,21 +140,36 @@ func (opr *OraclePriceRecord) Validate(c *config.Config, dbht int64) bool {
...
-	if opr.Version != OPRVersion {
+	if opr.Version != common.OPRVersion(net, int64(dbht)) {
 		return false // We only support this version
 	}
 
 	// Validate all the Assets exists
-	return opr.Assets.ContainsExactly(common.AllAssets)
+	switch opr.Version {
+	case 1:
+		if len(opr.WinPreviousOPR) != 10 {
```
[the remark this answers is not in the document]
WhoSoup: i think the length check there is fine for v1, but there's no length check for v2. also having 25 shorthashes is going to present a problem which i'll detail in a separate post

WhoSoup:
> ven using the chain is a bit complicated as the pegnet testnet chanid is the same on the main network and on the test network.

what happens right now if you boot to the wrong network/chain? if it doesn't exist, could you add a fail if it detects that an eblock in the chain with the same db sequence has a different hash? that would prevent someone from connecting to the wrong network and corrupting their database

WhoSoup:
An OPR with 25 previous winners will no longer fit inside of a 1 EC entry. The following OPR:
```json
{"coinbase":"FA00000000000000000000000000000000000000000000000000","dbht":200000,"winners":["0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000","0000000000000000"],"minerid":"something reasonable","assets":{"PNT":0.6046602879796196,"USD":0.9405090880450124,"EUR":0.6645600532184904,"JPY":0.4377141871869802,"GBP":0.4246374970712657,"CAD":0.6868230728671094,"CHF":0.06563701921747622,"INR":0.15651925473279124,"SGD":0.09696951891448456,"CNY":0.30091186058528707,"HKD":0.5152126285020654,"KRW":0.8136399609900968,"BRL":0.21426387258237492,"PHP":0.380657189299686,"MXN":0.31805817433032985,"XAU":0.4688898449024232,"XAG":0.28303415118044517,"XPD":0.29310185733681576,"XPT":0.6790846759202163,"XBT":0.21855305259276428,"ETH":0.20318687664732285,"LTC":0.360871416856906,"RVN":0.5706732760710226,"XBC":0.8624914374478864,"FCT":0.29311424455385804,"BNB":0.29708256355629153,"XLM":0.7525730355516119,"ADA":0.2065826619136986,"XMR":0.865335013001561,"DASH":0.6967191657466347,"ZEC":0.5238203060500009,"DCR":0.028303083325889995}}
```

is **1424** bytes. With Factom entry overhead it comes to 1450 bytes.

WhoSoup:
> This does double the EC cost yes. Do you feel 2ECs is too expensive?

not in a cost sense, more in an optics. we've been advertising mining as 1ec/entry and using that to attract miners

Emyrk:
> > This does double the EC cost yes. Do you feel 2ECs is too expensive?
> 
> not in a cost sense, more in an optics. we've been advertising mining as 1ec/entry and using that to attract miners. fwiw we can solve this easily by switching from byte-heavy json to protobuf for oprs
> 
> we would lose human readability in the factom explorer but the pegnet chain already has a quarter of a million entries in it and i question the ability of someone to manually grade and verify oprs. anyone interested in the values is more likely to use sites like factoshi
> 
> we'd also gain a performance boost during grading if we didn't have to decode future json entries

I've been using the local node explorer to verify records, so it can still be done on a block level very easily. I would like to maintain the human readability, and the overhead of marshaling is negligible. I was doing some benchmarks with binary marshaling in factomd of an entry vs json, and the json was twice as slow, but the results ended up being 127K json marshals a second. The hashrate for lxrhash is slower than that.

PaulBernier:
> 
> 
> > > This does double the EC cost yes. Do you feel 2ECs is too expensive?
> > 
> > 
> > not in a cost sense, more in an optics. we've been advertising mining as 1ec/entry and using that to attract miners. fwiw we can solve this easily by switching from byte-heavy json to protobuf for oprs
> > we would lose human readability in the factom explorer but the pegnet chain already has a quarter of a million entries in it and i question the ability of someone to manually grade and verify oprs. anyone interested in the values is more likely to use sites like factoshi
> > we'd also gain a performance boost during grading if we didn't have to decode future json entries
> 
> I've been using the local node explorer to verify records, so it can still be done on a block level very easily. I would like to maintain the human readability, and the overhead of marshaling is negligible. I was doing some benchmarks with binary marshaling in factomd of an entry vs json, and the json was twice as slow, but the results ended up being 127K json marshals a second. The hashrate for lxrhash is slower than that.

Thinking more about it I am of the same opinion as Who. I think it's pretty hard to justify doubling the basis cost of miners just for the convenience of debugging of a few developers. I also do like to be able to see the data quickly in the explorer, but that'd be selfish to ask the main participants of the network, the miners, to pay double price for that. Also it probably allows us to have even a few more bytes added in the future while still remaining at 1 EC.

To sum up, if there is a technical solution to remain at 1 EC, I find it hard to not put in the little extra work for the sake of miners. Protobuf or anything alike is fine to me. Ideally in a separate PR :)

WhoSoup:
Just noticed that this is including sqlite for some reason in addition to leveldb for a dataset that really doesn't need a relational dataset (it's just one table). 

The gorm SQLite package you're using just imports mattn/go-sqlite3, which uses cgo:
```
go-sqlite3 is cgo package. If you want to build your app using go-sqlite3, you need gcc. However, after you have built and installed go-sqlite3 with go install github.com/mattn/go-sqlite3 (which requires gcc), you can build your app without relying on gcc in future.
```

that means the average user wouldn't be able to build this project on windows without something like mingw, which is a huge pain in the ass install

Emyrk:
@WhoSoup I was using sqllite for the time series data that feeds factoshi.io. There are a few tables, all relating to the `node` api. I can look for a pure go solution? Or maybe make the `node` it's own binary so users can build the miner without importing this.

Emyrk:
> > > > This does double the EC cost yes. Do you feel 2ECs is too expensive?
> > > 
> > > 
> > > not in a cost sense, more in an optics. we've been advertising mining as 1ec/entry and using that to attract miners. fwiw we can solve this easily by switching from byte-heavy json to protobuf for oprs
> > > we would lose human readability in the factom explorer but the pegnet chain already has a quarter of a million entries in it and i question the ability of someone to manually grade and verify oprs. anyone interested in the values is more likely to use sites like factoshi
> > > we'd also gain a performance boost during grading if we didn't have to decode future json entries
> > 
> > 
> > I've been using the local node explorer to verify records, so it can still be done on a block level very easily. I would like to maintain the human readability, and the overhead of marshaling is negligible. I was doing some benchmarks with binary marshaling in factomd of an entry vs json, and the json was twice as slow, but the results ended up being 127K json marshals a second. The hashrate for lxrhash is slower than that.
> 
> Thinking more about it I am of the same opinion as Who. I think it's pretty hard to justify doubling the basis cost of miners just for the convenience of debugging of a few developers. I also do like to be able to see the data quickly in the explorer, but that'd be selfish to ask the main participants of the network, the miners, to pay double price for that. Also it probably allows us to have even a few more bytes added in the future while still remaining at 1 EC.
> 
> To sum up, if there is a technical solution to remain at 1 EC, I find it hard to not put in the little extra work for the sake of miners. Protobuf or anything alike is fine to me. Ideally in a separate PR :)

I wouldn't say it's doubling, as you should factor in the data source costs.  Can we determine what a protobuf would come too? I'm calculating 662 + 6 extid bytes not including the dbht, the minerid, or any field descriptors/overhead.

```
# D = self reported difficulty
# V = Version byte
# N = Nonce
# Prices = 8 decimal digits instead of 4 now.
Assets + Prices + winners + FA + D + V + N
  30*3 + (30*9) + 25*10   + 36 + 8 + 1 + 5
```

Protobuf will probably get us under 1EC? But it might be pretty close.

Emyrk:
> Just noticed that this is including sqlite for some reason in addition to leveldb for a dataset that really doesn't need a relational dataset (it's just one table).
> 
> The gorm SQLite package you're using just imports mattn/go-sqlite3, which uses cgo:
> 
> > go-sqlite3 is cgo package. If you want to build your app using go-sqlite3, you need gcc. However, after you have built and installed go-sqlite3 with go install github.com/mattn/go-sqlite3 (which requires gcc), you can build your app without relying on gcc in future.
> 
> that means the average user wouldn't be able to build this project on windows without something like mingw, which is a huge pain in the ass install. i'd vastly prefer a solution without sqlite (since we're loading leveldb anyway)

Ah I see the import comes from Gorm. That makes it difficult to remove. I think the best thing I could offer is to make `pegnet node` it's own binary, so the miner is a different binary than the node.

WhoSoup:
> I wouldn't say it's doubling, as you should factor in the data source costs. Can we determine what a protobuf would come too? I'm calculating 662 + 6 extid bytes not including the dbht, the minerid, or any field descriptors/overhead.

The Factom Overhead of the entry, which includes the ExtIDs is worst case:

```
header + chain + extid size  + nonce + difficulty + version
  1    +  32   +   1         +  1+8  +    1+8     +   1+1   
```

**54 bytes**


If we are using DIDs (which have a string length ~83 characters), every asset value always requires a full 8 bytes and also includes the **name** of the asset. That comes out to 954 bytes (900 proto + 54 overhead). 

However, since protobuf is ordered we can drop the asset names. We can also encode the prices as uint64. That brings it down to 728 + 54 bytes. 

Protobuf schema is:
```proto
syntax = "proto3";
package main;

message OPR {
    string address = 1;
    string id = 2;
    int32 height = 3;
    repeated string winners = 4;
    repeated Asset assets = 5;
}

message Asset {
    string name = 1;
    fixed64 value = 2;
}

message OPRMin {
    string address = 1;
    string id = 2;
    int32 height = 3;
    repeated string winners = 4;
    uint64 PEG = 5;
    uint64 pUSD = 6;
    uint64 pEUR = 7;
    uint64 pJPY = 8;
    uint64 pGBP = 9;
    uint64 pCAD = 10;
    uint64 pCHF = 11;
    uint64 pINR = 12;
    uint64 pSGD = 13;
    uint64 pCNY = 14;
    uint64 pHKD = 15;
    uint64 pKRW = 16;
    uint64 pBRL = 17;
    uint64 pPHP = 18;
    uint64 pMXN = 19;
    uint64 pXAU = 20;
    uint64 pXAG = 21;
    uint64 pXPD = 22;
    uint64 pXPT = 23;
    uint64 pXBT = 24;
    uint64 pETH = 25;
    uint64 pLTC = 26;
    uint64 pRVN = 27;
    uint64 pXBC = 28;
    uint64 pFCT = 29;
    uint64 pBNB = 30;
    uint64 pXLM = 31;
    uint64 pADA = 32;
    uint64 pXMR = 33;
    uint64 pDASH = 34;
    uint64 pZEC = 35;
    uint64 pDCR = 36;
}
```

Emyrk:
> Just noticed that this is including sqlite for some reason in addition to leveldb for a dataset that really doesn't need a relational dataset (it's just one table).
> 
> The gorm SQLite package you're using just imports mattn/go-sqlite3, which uses cgo:
> 
> > go-sqlite3 is cgo package. If you want to build your app using go-sqlite3, you need gcc. However, after you have built and installed go-sqlite3 with go install github.com/mattn/go-sqlite3 (which requires gcc), you can build your app without relying on gcc in future.
> 
> that means the average user wouldn't be able to build this project on windows without something like mingw, which is a huge pain in the ass install. i'd vastly prefer a solution without sqlite (since we're loading leveldb anyway)

Ok, I have pulled the node into here -> https://github.com/pegnet/pegnet-node
This removes any sqllite dependency from the core. It also allows me to be a bit loser with that development.

Emyrk:
> > I wouldn't say it's doubling, as you should factor in the data source costs. Can we determine what a protobuf would come too? I'm calculating 662 + 6 extid bytes not including the dbht, the minerid, or any field descriptors/overhead.
> 
> The Factom Overhead of the entry, which includes the ExtIDs is worst case:
> 
> ```
> header + chain + extid size  + nonce + difficulty + version
>   1    +  32   +   1         +  1+8  +    1+8     +   1+1   
> ```
> 
> **54 bytes**
> 
> If we are using DIDs (which have a string length ~83 characters), every asset value always requires a full 8 bytes and also includes the **name** of the asset. That comes out to 954 bytes (900 proto + 54 overhead).
> 
> However, since protobuf is ordered we can drop the asset names. We can also encode the prices as uint64. That brings it down to 728 + 54 bytes.
> 
> Protobuf schema is:
> 
> ```protobuf
> syntax = "proto3";
> package main;
> 
> message OPR {
>     string address = 1;
>     string id = 2;
>     int32 height = 3;
>     repeated string winners = 4;
>     repeated Asset assets = 5;
> }
> 
> message Asset {
>     string name = 1;
>     fixed64 value = 2;
> }
> 
> message OPRMin {
>     string address = 1;
>     string id = 2;
>     int32 height = 3;
>     repeated string winners = 4;
>     uint64 PEG = 5;
>     uint64 pUSD = 6;
>     uint64 pEUR = 7;
>     uint64 pJPY = 8;
>     uint64 pGBP = 9;
>     uint64 pCAD = 10;
>     uint64 pCHF = 11;
>     uint64 pINR = 12;
>     uint64 pSGD = 13;
>     uint64 pCNY = 14;
>     uint64 pHKD = 15;
>     uint64 pKRW = 16;
>     uint64 pBRL = 17;
>     uint64 pPHP = 18;
>     uint64 pMXN = 19;
>     uint64 pXAU = 20;
>     uint64 pXAG = 21;
>     uint64 pXPD = 22;
>     uint64 pXPT = 23;
>     uint64 pXBT = 24;
>     uint64 pETH = 25;
>     uint64 pLTC = 26;
>     uint64 pRVN = 27;
>     uint64 pXBC = 28;
>     uint64 pFCT = 29;
>     uint64 pBNB = 30;
>     uint64 pXLM = 31;
>     uint64 pADA = 32;
>     uint64 pXMR = 33;
>     uint64 pDASH = 34;
>     uint64 pZEC = 35;
>     uint64 pDCR = 36;
> }
> ```

If we drop the assets on the json, most users will be below the 1K limit and we can postpone a protobuf encoding until we get DIDs. I'll begin implementing a protobuf encoding though, just throwing that out there.

Emyrk:
@PaulBernier @WhoSoup Can we push the protobuf to a pr after this? Can we merge this in, get the rename branch in, then get a protobuf pr in?

Issue #270, opened 2019-08-24
Change ticker from PNT to PEG


DavidChapman77:
PNT is taken by https://coinmarketcap.com/currencies/penta/.  Based upon my research, the ticker PEG is available.

PNT needs to be changed in all code, documentation, and websites.

stackdump:
Changed on Whitpaper.md and Mining Wiki

PR #275, opened 2019-08-27
Pnt 270 pnt rename

Emyrk:
PNT -> PEG rename

The code will always handle PEG, and the json marshaling will do the switch

Issue #270


Emyrk:
Rebased https://github.com/pegnet/pegnet/pull/281

PR #276, opened 2019-08-28
PNT-193 : Rate Calculation

Emyrk:
**Using Fixed Points**
The main function implemented added only calculate the rates if we convert. I did not implement conversions on chain.  
`Exchange` -> Fixed input conversion  

1.  Choose 2 random usd rates for from and to
2.  Choose a random amount of from tokens (this is have)
3.  Convert those tokens to to for an amt
4.  Given that amount, how many from tokens does it take to get amt of to. (this is need`).
5.  Compare need and have

The unit test in table format:
```markdown
| TX | From Currency (int64) | Conversion Direction (float64 ratio) | To Currency (int64) |
|----|-----------------------|:------------------------------------:|---------------------|
| 1  | Given `have`          |                 --->                 | Computed `amt`      |
| 2  | Computed `need`       |                 --->                 | Given `amt`          |
| 3  | Computed `whenNeeded` |                 <---                 | Given `amt`         |
```

Results: `need`  and `have` do not match in every case. `whenNeeded` and `need` do match (they can be off by 1 because of truncation, but only 1). This is because we lose precision going from `have` to `amt`, and then any math with `amt` will carry that precision error. I made sure to only truncate, meaning the lost precision will result in lost tokens, not gained. And this loss will be very small. This is an inherit problem we have with conversion ratios. For example: 

If BTC is at 10K USD, and FCT at $5, then:
1 satoshi is $0.0001 and 1 factoshi is $0.00000005, then you need a minimum 2000 factoshis to get 1 satoshi. If you transact 1999 factoshis, the current code will result in 0 satoshis. Rounding up would create tokens, rounding down destroys tokens in the tx. If we rounded up, I think there would be some office space type vulnerability, where if the amount generated rounding up is more than the cost of 1 EC, it would be beneficial to make those txs.


Emyrk:
Being move to a module

PR #281, opened 2019-09-04
PNT-270:  pnt rename rebased

Emyrk:
PNT -> PEG rename

The code will always handle PEG, and the json marshaling will do the switch

Issue #270


opr/grader.go
```diff
@@ -304,6 +303,11 @@ func (g *QuickGrader) Sync() error {
 }
 
 func (g *QuickGrader) FetchOPRBlock(block *EntryBlockMarker) (*OprBlock, error) {
+	// We want to set the OPR version for the network. This is kinda hacky, but the marshal json
```
PaulBernier: Agree with the hacky command 😄 It probably means there is a design problem somewhere. But not something to fix in this PR anyway

opr/opr.go
```diff
@@ -427,9 +429,78 @@ func (opr *OraclePriceRecord) CreateOPREntry(nonce []byte, difficulty uint64) (*
...
+	if opr.Version == 0 {
+		return fmt.Errorf("opr version is 0")
+	}
+
+	err := json.Unmarshal(data, opr)
+	if err != nil {
+		return err
+	}
+
+	// If version 1, we need to swap PNT and PEG
```
PaulBernier: Can you explain why the asset needs to be swapped directly inside the OPR? I am not convinced we should alter directly the OPR retroactively from the inside... you don't delete XPD/XPT for instance

opr/assetList.go
```diff
@@ -44,10 +45,22 @@ func (o OraclePriceRecordAssetList) List(version uint8) []Token {
 
 // from https://github.com/iancoleman/orderedmap/blob/master/orderedmap.go#L310
 func (o OraclePriceRecordAssetList) MarshalJSON() ([]byte, error) {
+	if _, ok := o["version"]; !ok {
+		return nil, fmt.Errorf("marshaling json must be called through a safe function")
+	}
+	version := int(o["version"])
+	delete(o, "version")
```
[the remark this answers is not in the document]
Emyrk: So in v1 we use the `json.Marshal(opr)`, but there is a problem. In the codebase, I switch everything to PEG. Rather than putting if statements everywhere on when to use `PNT` and when to use `PEG`, I thought it would be easier to use `PEG` everywhere. When we go to marshal the content (which is signed), we switch `PEG->PNT`. So the only place we ever see `PNT` is in the marshaled content. No code will ever use `PNT`, as when we unmarshal, we again replace `PNT` with `PEG` in the map.


The `version` thing is for 2 reasons. Calling `json.Marshal` is inherently unsafe, as now we have this rename business we ALWAYS need to address when marshaling content. The hashmap is already a custom marshal function that makes sure to marshal exactly the right assets, and nothing more.

The annoying part is the hashmap for all the currencies has no access to the OPR it is apart of, meaning it has no idea what `version` it is for. Because this is a `json.Marshal` override function, we cannot adjust the method definition without having to write our own Json encoding library (no thanks...). So I use an index in the hashmap as an indicator of which version to marshal for. It was that or creating a new structure, but creating a new structure would require touching almost every file (a lot touch the assets). In the protobuf rename, I am moving to use accessors vs using the map directly, partly because of this reason. Dealing with the raw map limits how I can add context to it (like a version number).

opr/grader.go
Emyrk: Yea, I ended up passing context to the marshal json function via another route. I will adjust this comment, as it's still used in the polling library. We can remove this once v2 is activated

opr/opr.go
Emyrk: XPD and XPT are valid v1 assets. The code internally only deals with `PEG`, there is no reference to `PNT` except in the marshaling functions. This means all v1 users when upgrading will query their `PEG` balance, not their `PNT`.

If I did not do this, everyone would have a `PNT` and `PEG` balance, and I think it would add more confusion than it is worth. We are stuck with `PNT` on chain, and that is the only place it exists.

opr/grader.go
Emyrk: Oh I lied. I actually passed the OPRVersion to the polling by adjusting the method definition. I can remove this and that global. Forgot to clean that up :+1:

Emyrk:
> The translation/override of PNT to PEG in the OPR structure is a bit too convoluted and makes things pretty hard to follow... I hope we can clean up all this very soon and meanwhile you don't get lost with that logic...

We can either have the code maintain PNT and PEG, or just PEG. I decided to just maintain PEG. The only time we need PNT is when we write to chain, so it only needs to occur when we marshal the content.

PaulBernier:
> > The translation/override of PNT to PEG in the OPR structure is a bit too convoluted and makes things pretty hard to follow... I hope we can clean up all this very soon and meanwhile you don't get lost with that logic...
> 
> We can either have the code maintain PNT and PEG, or just PEG. I decided to just maintain PEG. The only time we need PNT is when we write to chain, so it only needs to occur when we marshal the content.

I understand. But the amount of side effects (mutate objects on the fly in multiple places) scares me, I personally embrace the "immutable objects everywhere" motto as much as possible nowadays, and it works well for me. Anyway I approved the PR.

PR #282, opened 2019-09-04
Pnt 280 : Use protobufs vs json for opr content

Emyrk:
# Switch to Profotbufs

Entry content (oprs) is now protobuf encoded. In order to maintain the ability to debug, I added 2 cmd line commands:

```
pegnet decode eblock 25e22578064724ab3d96fee1cf5048f4bde740c4a50169def07065b027644ea6  
pegnet decode entry 1af4ca48aaafedaea9ba1fd0dc541c88c4e2d833e7f4b34bc1b95576e9feee86
```

You can pipe it through `jq`, all output is json (or an error).

# Float64 -> uint64

The protobuf encoding is in uint64, so all price quotes are multiplied by 1e8, meaning we allow precision to the 8th decimal point. Because the old precision was only 4 decimal points, this is backward compatible. This also means we should not have to truncate anywhere in the code, as adding/removing to the map will handle it.

In order to not have to rewrite a lot of code, there are still float accessors. I marked them as deprecated, as they should be avoided, and we should switch to only using fixed point math everywhere. 

# Extra Validation Rule

@PaulBernier pointed out enforcing the minerid is a new validation rule. At the time I responded, the only miner in use was the reference miner. This may no longer be true, so I added a conditional so that check is only in v2


opr/oprencoding/opr.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+package oprencoding;
+
+message ProtoOPRMin {
```
PaulBernier: I don't think the Min suffix is necessary anymore

opr/oprencoding/opr.proto
```diff
@@ -0,0 +1,39 @@
...
+    string address = 1;
+    string id = 2;
+    int32 height = 3;
+    repeated string winners = 4;
+    uint64 PEG = 5;
```
PaulBernier: Why not `repeated uint64 assets = 5? Then have the order fixed somewhere and read/write that array accordingly? It could even takes slightly less space isn't?

opr/oprencoding/opr.proto
Emyrk: I can drop the Min.

opr/oprencoding/opr.proto
Emyrk: There were some talks about that, but we wanted the asset names attached to the price. Doing it like these saves space vs an array of `{"Name":"PEG", "Price":0}`, but still keeps the asset names attached to the price. 

It makes the code a bit more tedious, and with some reflect magic it would not be that hard to iterate over. I imagine in JS it could even be simpler. Maybe we should nest the fields in a sub class/struct?

Emyrk: I just tested it, currently, with a random id I choose of length 16, the entry is 893 bytes. With an array of uint64, it is 844 bytes.

If we do an array of {Name, Price}, it is 1082 bytes.
Doing a map[string]uint64 is 1082 bytes.

Emyrk:
Switched from winners from []strings to [][]byte, now down to `693` bytes for the entry.

Also added a more involved unit test for marshal checking

opr/assetList.go
```diff
@@ -29,6 +30,28 @@ func (o OraclePriceRecordAssetList) ContainsExactly(list []string) bool {
...
+}
+
+// Uint64Value returns 0 for empty assets
+func (o OraclePriceRecordAssetList) Uint64Value(asset string) uint64 {
+	return o[asset]
+}
+
+// Value converts the value to a float
+// Deprecated: Should not be using floats, stick to the uint64
+func (o OraclePriceRecordAssetList) Value(asset string) float64 {
```
PaulBernier: Do I understand correctly that for v1 this is the value that ends up in the OPR after transformation:
```
float64(uint64(TruncateTo4Decimales(value) * 1e8)) / 1e8
```

opr/oprencoding/opr.proto
WhoSoup: > Why not `repeated uint64 assets = 5? Then have the order fixed somewhere and read/write that array accordingly? It could even takes slightly less space isn't?

the only advantage i can think of is that it's a self-describing message that other languages / tools can import without explicitly also needing the list of assets. 

any change to the assets is going to require a new opr version and activation height anyway, so i think making it `repeated uint64` is fine from that perspective

i don't think we can fit the name of the asset into the message and we shouldn't

opr/assetList.go
```diff
@@ -38,11 +61,24 @@ func (o OraclePriceRecordAssetList) List(version uint8) []Token {
 	tokens := make([]Token, len(assets))
 	for i, asset := range assets {
 		tokens[i].Code = asset
-		tokens[i].Value = o[asset]
+		tokens[i].Value = o.Value(asset)
```
PaulBernier: And here you are again converting your uint64 to float64. So I don't get the motivation to use uint64 as you are literally always converting it to float64

PaulBernier: If you are not using uint64, then I suggest to stick with float64, and later have a V3 that is actually using uint64 everywhere. I don't see the benefit of doing it halfway

WhoSoup: > If you are not using uint64, then I suggest to stick with float64, and later have a V3 that is actually using uint64 everywhere. I don't see the benefit of doing it halfway

the deciding factor here is actually the format we get from the APIs. there's no math involved in writing OPRs to the chain, it's just a 1:1 transfer from the API to the record. 

our apis are providing the exchange rate as float64 as far as i can tell (i checked 1forge, cmc, currencylayer), so it makes sense to keep them as float64 in the opr itself.

it's only when we start doing math (both addition and multiplication) with them that it matters we use some sort of fixed point or whole number system in order to improve accuracy. 

so i suppose the question is: do we need improved accuracy during the grading process? i don't think so, since the results of the grade are not used for anything so any float error doesn't compound or has any impact 

converting them to a fixed format in the opr saves us some processing power later but that's probably negligible

Emyrk: @WhoSoup Explained it perfectly. I still think it'd be beneficial to move away from floats for the grading too, but at this time, I don't see a huge pressure too. With conversions on the docket, I want to ensure the rates put on the chain are the same rates pulled and used.

opr/assetList.go
Emyrk: Yes. 

V1 stores 4 decimal place, and this is crucial. If we did not truncate it, the trailing oddities of floats would be in the uint64, and the value put in might not come out the same.

PaulBernier: What I meant is that in this new code you go through all those transformations `float64(uint64(TruncateTo4Decimales(value) * 1e8)) / 1e8` while in the old code we were using directly `TruncateTo4Decimales(value)` for OPR V1. What I am wondering if after all those successive transformations (i.e. floating point approximations) we always end up with the exact same result as with the old code. Otherwise it's a potential fork while V1 is still operating.

PaulBernier: Steven showed me a snippet of code that verifies the equality for a convincing number of examples. So it does look fine.

Emyrk: https://play.golang.org/p/qyRUgPFfo41

Here is the process

opr/oprencoding/opr.proto
sambarnes: Think I'm in agreement about using the uint64 array rather than a map. We already have forced ordering for grading so as long as this array shares that same order, I don't think there is much of a downside. Staying under 1 EC was the primary motivation for moving to protobufs anyways, so it seems a little wasted if we used the map that pushes us back into >1 EC territory. 

Also, a minor benefit is that if (when) PEG ticker is changed again, we won't need to modify the OPR code again to accommodate.

Emyrk: @WhoSoup @PaulBernier Hey guys, we have like ~750 bytes with a long DID, and will probably release a new proto spec if we update. Why not just use a map? It was 1082 before we dropped 25*8 bytes for the winners, and keep the proto def `Asset = Price` until we need to bring it under 1 EC and be smart again?

PaulBernier: > @WhoSoup @PaulBernier Hey guys, we have like ~750 bytes with a long DID, and will probably release a new proto spec if we update. Why not just use a map? It was 1082 before we dropped 25*8 bytes for the winners, and keep the proto def `Asset = Price` until we need to bring it under 1 EC and be smart again?

"Influencers" always say "don't compromise" :smile: So I think we should go with `repeated uint64 assets = 5` now rather than just postponing and relying on our future self being smarter

Emyrk: I'll go with `repeated uint64 assets = 5`

Emyrk:
@WhoSoup @PaulBernier I switched it to `[]uint64`

https://github.com/pegnet/pegnet/pull/282/commits/b44cf5e0e198d192b1696fea45d835c6f40195d5

PaulBernier:
Did my own testing against this implementation and it seems to work as expected, approving.

opr/assetList.go
```diff
@@ -71,9 +107,8 @@ func (o OraclePriceRecordAssetList) MarshalJSON() ([]byte, error) {
 		// add key
 		kEscaped := strings.Replace(k, `"`, `\"`, -1)
 		s = s + `"` + kEscaped + `":`
-		// add value
-		v := o[k]
-		vBytes, err := json.Marshal(v)
+		// add value as float.
+		vBytes, err := json.Marshal(o.Value(k))
```
WhoSoup: in v2 we're saving the value as uint64 on chain but here it's encoding the v2 json as a float.

WhoSoup:
the way we're using uint64 doesn't work for big numbers (specifically numbers close to MaxUint64 and above), see: https://play.golang.org/p/yED9f55Ud71

A number that big is "reasonable", since 184,467,440,737 KRW ~= 150 million USD. Can we weigh the likelihood of something like this occurring? We'd have to build safety checks into wallets to ensure that users are aware a single address can't hold that many won and while doing transactions and conversions, we have to check for overflows.

the alternative is using something like bigInt instead of uint64

WhoSoup:
switching to uint64 for the file format is a massive headache for implementation with no practical gain.

the apis all provide data in float64.
v1 writes float64 to chain, v2 converts float64 to uint64 and writes uint64 to chain
when parsing, v1 gets converted to uint64, v2 reads directly
in order to grade, the uint64 gets converted back to float64

that means for grading, we're taking a float, converting it to uint64, then converting it back to a float at some point. no gain.

the benefit of using non-floats only comes after the opr leaves the opr and grading module but not inside those modules. therefore it makes much more sense to just keep it consistently float64 and only convert it to another format when leaving OPRs/grading to do transactions/balance/wallets

basically the only opr whose values ever need to be converted from float64 to uint64 is the winning OPR's values

Emyrk:
> the way we're using uint64 doesn't work for big numbers (specifically numbers close to MaxUint64 and above), see: https://play.golang.org/p/yED9f55Ud71
> 
> A number that big is "reasonable", since 184,467,440,737 KRW ~= 150 million USD. Can we weigh the likelihood of something like this occurring? We'd have to build safety checks into wallets to ensure that users are aware a single address can't hold that many won and while doing transactions and conversions, we have to check for overflows.
> 
> the alternative is using something like bigInt instead of uint64

We do have to deal with overflows, I agree and actually brought this up to Luap. The assetlist is not balances, it is prices. So using uint64s for prices means as long as no currency goes above `2^66` USD, we will not overflow. So uint64s for price quotes is safe.

> switching to uint64 for the file format is a massive headache for implementation with no practical gain.
I disagree. Ignoring grading for the moment, and just for the conversions. Conversions use uint64, so we need to go float -> int at some point before conversions. We either do that before we go on-chain, or after. If we do it before, everyone is guaranteed the same int when they parse the opr. This removes any floats from the conversion calculation, as the value on-chain is in the right format.

For grading. Currently, we go float -> uint64 -> float for grading. I agree this is not ideal, and I'd like to drop floats from grading later.

PaulBernier:
>  We either do that before we go on-chain, or after. If we do it before, everyone is guaranteed the same int when they parse the opr.

The conversion uint64->float64 is as safe as the conversion float64->uint64, so the guarantee seems pretty much the same whether the conversion happens before or after being stored on chain, isn't it? And if I follow Who is coming to the same conclusion that this result in the code being over convoluted because of back and forth conversions <https://github.com/pegnet/pegnet/pull/282#discussion_r321983635>.

Emyrk:
There was a lot of pushback about using floats, I was starting to get us to move away from them. I prefer to store the actual value that will be used in conversions on-chain, rather than something that needs to be converted.

Are we in agreeance that it would be ideal if grading could use the uint64 vs floats, and drop floats entirely from pegnet?

WhoSoup:
i don't think floats should be used for balances, transactions, or conversions for obvious reasons. uint64 is a compromise but ideally we should use big-ints (doesn't necessarily have to be golang's implementation, we can look for finance libraries that provide more useful ones)

for OPRs themselves however, there's zero difference between float64 and uint64. they're interchangeable within the constraints of the data we have. currently, grading uses float64, meaning every opr has to provide the values in float64 at some point. there is only one OPR every block whose values will be used in balances/transactions/conversions, and that's the block's winner

if we use uint64 during the grading process, then that's a reason to make oprs write them to chain, but in v2 right now we don't. if they provided more accuracy than floats then that's also a reason to write them to chain but they don't.

if we saved processing power by converting them to uint64 "client-side" and not have to do it during grading then that's a reason but we're just converting them back to float so we're wasting processing power.

Emyrk:
> i don't think floats should be used for balances, transactions, or conversions for obvious reasons. uint64 is a compromise but ideally we should use big-ints (doesn't necessarily have to be golang's implementation, we can look for finance libraries that provide more useful ones)

Agreed. 

> for OPRs themselves however, there's zero difference between float64 and uint64. they're interchangeable within the constraints of the data we have. currently, grading uses float64, meaning every opr has to provide the values in float64 at some point. there is only one OPR every block whose values will be used in balances/transactions/conversions, and that's the block's winner
>
> if we use uint64 during the grading process, then that's a reason to make oprs write them to chain, but in v2 right now we don't. if they provided more accuracy than floats then that's also a reason to write them to chain but they don't.

In v2 we will use uint64. v2 has conversions in it, and currently, the implementation of rates uses the GoLang's BigInts. They require an int64 type. So conversion from uint64 -> int64 will happen. If we keep it as a float, then the conversion from float -> int64 will happen. When you convert float -> int/uint, if you want the process to be symmetrical, you have to round.

https://play.golang.org/p/JrUWqbOU96K

If you do not use this snippet, then going backwards yields a different float. So if we ever go float -> int, all developers have to agree exactly how to do this. As multiple implementations (a truncate vs a round vs doing nothing) would yield small differences. (Keep in mind for the miners, they can go float-> uint64 however they would like, as 1e-8 is barely a difference for a price quote, but means a lot when we talk about balances)
```
func ToUint64R(v float64) uint64 {
	return uint64(math.Round(v * 1e8))
}
```

Going int -> float is much more intuitive, as there is no rounding or truncating. In my mind, this means it is safer for another library to implement int -> float, but it can be unsafe to go float -> int.
```
func ToFloat64(u uint64) float64 {
	return float64(u) / 1e8
}
```

So we if store uint64 on-chain, there is only 1 conversion. int -> float for grading. (__And ideally, this would be phased out to remove floats entirely.__)

If we store float on-chain, there is only 1 conversion, float -> int for conversions. We now have to ensure the conversion is done identically in every wallet or tool that interacts with any conversion.

> if we saved processing power by converting them to uint64 "client-side" and not have to do it during grading then that's a reason but we're just converting them back to float so we're wasting processing power.

I don't think you could even see the conversion on a profiler. A quick benchmark of the ToUint with the rounding yields:
```
BenchmarkToUint64R-8   	1000000000	         2.54 ns/op
```

---------------------

Sidenote, the proto encoding of double is a `double` in java, and a `float64` in golang. 
<https://developers.google.com/protocol-buffers/docs/proto#scalar>

Would that yield a different result if we used a double in one language and a float in another?

opr/assetList.go
Emyrk: Meaning encode them as uint64s?

WhoSoup:
we're settling on using uint64 in the v2 opr. the pro/cons are pretty close but what tipped me the over the edge was that we're going to disallow submitting negative values in v2 anyway so we might as well use a format that doesn't allow negative numbers to prevent any accidents

also i think with varint, a uint64 will use less space than a float64

Emyrk:
For posterity. Only the Rounding method is symmetrical

https://play.golang.org/p/UxxkgO0oOhR

opr/assetList.go
WhoSoup: yes

Emyrk: @WhoSoup After some conversation, are you ok with keeping this a float value in the JSON?

PR #308, opened 2019-09-16
PNT-294: stale polling

Emyrk:
This makes stale polling results fallback to the next data source. Once we find a datasource that is not stale (or we run out), we take the most recent quote out of our data sources. This would allow us to re-enable crypto on 1 forge, as that error they had awhile back was a stale price quote. 

So questions for reviewers before I make this a not-draft pr:

- Should I re-enable crypto on 1 forge in this PR?
- ~~Should we explicitly handle weekends for forex? (I think this is a must, just want to talk about how)~~
  - I ended up programming in the Open/Close market times for Forex and Commodities. Holidays and such will count as "open" and all data sources will be queried. So that means if you have 4 sources for forex, all 4 will be queried when the market is closed on an unusual time. Do you think we will need a syntax for "only fallback during a price error, not a stale error"?
- Is 30 min a good "stale" duration?

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -108,6 +108,7 @@ func NewUnitTestConfigProvider() *UnitTestConfigProvider {
   OpenExchangeRatesKey=CHANGEME
   CoinMarketCapKey=CHANGEME
   1ForgeKey=CHANGEME
+  StaleQuoteDuration=10m
 
 
 [OracleDataSources]
--- a/common/config.go
+++ b/common/config.go
@@ -29,6 +29,10 @@ const (
 
 	ConfigCoinMarketCapKey = "Oracle.CoinMarketCapKey"
 	Config1ForgeKey        = "Oracle.1ForgeKey"
+
+	// ConfigStaleDuration determines how old a quote is allowed to be and still be
+	// acceptable
+	ConfigStaleDuration = "Oracle.StaleQuoteDuration"
 )
 
 // DefaultConfigOptions gives us the ability to add configurable settings that really
@@ -53,12 +57,13 @@ func (c *DefaultConfigOptions) Load() (map[string]string, error) {
 	settings[ConfigMinerDBType] = "ldb"
 	settings[ConfigPegnetNodeDBPath] = "$PEGNETHOME/data_$PEGNETNETWORK/node.sqlite"
 	settings[ConfigControlPanelPort] = "8080"
+	settings[ConfigStaleDuration] = "10m"
 
 	return settings, nil
 }
 
 func NewUnitTestConfig() *config.Config {
-	return config.NewConfig([]config.Provider{NewUnitTestConfigProvider()})
+	return config.NewConfig([]config.Provider{NewDefaultConfigOptionsProvider(), NewUnitTestConfigProvider()})
 }
 
 // UnitTestConfigProvider is only used in unit tests.
--- a/common/config.go
+++ b/common/config.go
@@ -57,7 +57,7 @@ func (c *DefaultConfigOptions) Load() (map[string]string, error) {
 	settings[ConfigMinerDBType] = "ldb"
 	settings[ConfigPegnetNodeDBPath] = "$PEGNETHOME/data_$PEGNETNETWORK/node.sqlite"
 	settings[ConfigControlPanelPort] = "8080"
-	settings[ConfigStaleDuration] = "10m"
+	settings[ConfigStaleDuration] = "30m"
 
 	return settings, nil
 }
--- a/polling/assets.go
+++ b/polling/assets.go
@@ -332,10 +332,12 @@ func (d *DataSources) PullBestPrice(asset string, reference time.Time, sources m
 			prices = append(prices, price)
 
 			// We can break out if this is a 'good' price
-			//	Is it stale?
-			if reference.Sub(price.When) > d.staleDuration {
+			//	Is it stale AND the market is open? We can expect stale quotes in a closed market
+			if reference.Sub(price.When) > d.staleDuration && IsMarketOpen(asset, reference) {
 				// This price quote is stale, keep fetching prices
-				continue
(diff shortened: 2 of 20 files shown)


common/config.go
```diff
@@ -103,6 +108,7 @@ func NewUnitTestConfigProvider() *UnitTestConfigProvider {
   OpenExchangeRatesKey=CHANGEME
   CoinMarketCapKey=CHANGEME
   1ForgeKey=CHANGEME
+  StaleQuoteDuration=10m
```
Emyrk: Unit tests have a 10m stale default

polling/assets.go
```diff
@@ -298,6 +303,65 @@ func (d *DataSources) PullAllPEGAssets(oprversion uint8) (pa PegAssets, err erro
...
+		}
+	}
+
+	// No prices found, and the last data source had an error
+	if err != nil && len(prices) == 0 {
+		return
+	}
+
+	// pa.When is uninitialized, so the reference.Sub(pa.When) == a very large value
+	mostRecent := reference.Sub(pa.When)
```
WhoSoup: this is an unnecessarily confusing way of doing it, you don't need 'mostrecent' at all, you can just do:

```go
	for _, price := range prices {
		if price.When > pa.When {
			pa = price
		}
	}
```

polling/assets.go
```diff
@@ -298,6 +303,65 @@ func (d *DataSources) PullAllPEGAssets(oprversion uint8) (pa PegAssets, err erro
 	return pa, nil
 }
 
+// PullBestPrice pulls the best asset price we can find for a given asset.
+// Params:
+//		asset		Asset to pull pricing data
+//		reference	Time reference to determine 'staleness' from
+//		sources		Map of datasources to pull the price quote from.
+func (d *DataSources) PullBestPrice(asset string, reference time.Time, sources map[string]IDataSource) (pa PegItem, err error) {
```
WhoSoup: the way this works at the moment is:
1. it goes through the apis by the priority order
2. if it encounters a result that is not stale, it will return that
3. if all responses are stale, it returns the most recent one

1 and 2 work (albeit with some changes below), but i had a question for 3: if all api sources are stale, should we give them the newest, or abide by the intended configuration? you could make the argument that if all sources are stale there's likely something wrong (ie markets closed unexpectedly) and we should serve them with the stale data from the api they configured

polling/assets.go
```diff
@@ -298,6 +303,65 @@ func (d *DataSources) PullAllPEGAssets(oprversion uint8) (pa PegAssets, err erro
...
+			prices = append(prices, price)
+
+			// We can break out if this is a 'good' price
+			//	Is it stale AND the market is open? We can expect stale quotes in a closed market
+			if reference.Sub(price.When) > d.staleDuration && IsMarketOpen(asset, reference) {
+				// This price quote is stale, keep fetching prices
+				continue
+			}
+
+			// This price is acceptable
```
WhoSoup: if we find an acceptable price, can we just return it here right away? if the code gets to this point, the only price item that is not stale is the last item in the slice, it is therefore automatically the most recent

polling/assets.go
Emyrk: So if all sources are stale, return the highest priority one? I can make that change, that does seem reasonable.

polling/assets.go
Emyrk: That is true, I can just return here yes

polling/assets.go
```diff
@@ -298,6 +303,65 @@ func (d *DataSources) PullAllPEGAssets(oprversion uint8) (pa PegAssets, err erro
...
+			break
+		}
+	}
+
+	// No prices found, and the last data source had an error
+	if err != nil && len(prices) == 0 {
```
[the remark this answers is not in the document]
Emyrk: Currently, the polling can return a 0 price. If you, for example, had 0 data sources, the polling would not fail. There would be no error, and the list length is 0.

The opr creation would fail because a 0 price is an invalid opr. I do not want to enforce a non-zero in the polling, but rather do that in the opr validation. For example, `PEG` goes through this process, but returns a 0.

polling/assets.go
Emyrk: Good point :+1:

polling/assets.go
```diff
@@ -298,6 +303,68 @@ func (d *DataSources) PullAllPEGAssets(oprversion uint8) (pa PegAssets, err erro
...
+		return
+	}
+
+	// If we got here, that means that there might exist some price quotes from
+	// our data sources, but they are all stale. Instead of taking the most recent price
+	// quote, we will take the highest priority quote given our data-source order.
+	if len(prices) > 0 {
+		pa = prices[0]
+		return pa, nil
+	}
```
Emyrk: @WhoSoup The highest priority price quote is now taken if all quotes are stale.

polling/assets.go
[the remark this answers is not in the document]
WhoSoup: I think a cleaner solution would be to have a non-configurable FixedPEG source that return 0, like FixedUSD, or just not calling a data source for peg/usd, rather than hiding the implicit value of PEG = 0 in a hard to understand if-statement

a quick solution would be to just put an explicit `if asset == "PEG" { return }` before the loop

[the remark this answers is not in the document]
WhoSoup: would this also just return a 0 if someone turned off all apis for a specific asset in the config?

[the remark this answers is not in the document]
Emyrk: Looking back, I should add FixedPEG, but that would require a config update. I'll create a fixed PEG, and add it to the new config, and at some point we can enforce it going forward.

I will add an `if asset == "PEG" { return }` for now, and make an error if the other sources are not configured. At some point in the future, we can drop that and people will have to turn on the `FixedPEG` like they do the `FixedUSD`

Emyrk:
@WhoSoup Can we merge this?

PR #315, opened 2019-09-19
Added various unit tests

Emyrk:
Found some slight bugs in the parse.
- The order of V1 -> V2 is important for general parsing. I know this is more expensive as v2 is the only case going forward, but we should use the correct parse function since we know the version from the extids.
- The parsing was not returning errors in some cases (like []byte{})
- The parser had a bug in the error return (it was not returning the error)

The rest is just unit tests


modules/grader/base.go
```diff
@@ -2,8 +2,31 @@ package grader
...
+	lxInitializer.Do(func() {
+		// This code will only be executed ONCE, no matter how often you call it
+		if size, err := strconv.Atoi(os.Getenv("LXRBITSIZE")); err == nil && size >= 8 && size <= 30 {
+			LX = lxr.Init(lxr.Seed, uint64(size), lxr.HashSize, lxr.Passes)
+		} else {
+			LX = lxr.Init(lxr.Seed, lxr.MapSizeBits, lxr.HashSize, lxr.Passes)
+		}
+
+	})
+}
```
Emyrk: This is how we did it before. It's not a bad way to handle it, but still feels a bit off

modules/factoidaddress/factoidaddress.go
```diff
@@ -0,0 +1,51 @@
+package factoidaddress
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/sha256"
+	"fmt"
+)
+
+var FactoidAddressPrefix = [2]byte{0x5f, 0xb1}
```
WhoSoup: i'd prefer making it a slice, no reason for it to be an array, just makes the code more confusing later on

modules/factoidaddress/factoidaddress.go
```diff
@@ -0,0 +1,51 @@
...
+	if len(data) != 2+32+4 {
+		return fmt.Errorf("address is of wrong length")
+	}
+
+	prefix := data[:2]
+	if bytes.Compare(prefix, FactoidAddressPrefix[:]) != 0 {
+		return fmt.Errorf("address has wrong prefix")
+	}
+
+	checksum := data[len(data)-4:]
```
WhoSoup: since we're making this a module, we should factor out checksum() into its own function

modules/factoidaddress/factoidaddress.go
```diff
@@ -0,0 +1,51 @@
...
+	shad := sha256.Sum256(sha[:])
+
+	if bytes.Compare(shad[:4], checksum) != 0 {
+		return fmt.Errorf("checksum is not correct")
+	}
+
+	return nil
+}
+
+func Random() string {
```
WhoSoup: instead of having a Random(), can we just turn this into an "Encode(payload []byte) string" that takes the given payload and prepends the prefix and appends the checksum? then this function has a lot more use 

`Random()` could then just be:
```go
func Random() string {
    addr := make([]byte, 32)
    rand.Read(addr)
    return Encode(addr)
}
```

modules/grader/base.go
WhoSoup: with the singletons, we don't need the `once` functionality anymore. just:
```go
var LX *lxr.LXRHash
func InitLX() {
	if size, err := strconv.Atoi(os.Getenv("LXRBITSIZE")); err == nil && size >= 8 && size <= 30 {
		LX = lxr.Init(lxr.Seed, uint64(size), lxr.HashSize, lxr.Passes)
	} else {
		LX = lxr.Init(lxr.Seed, lxr.MapSizeBits, lxr.HashSize, lxr.Passes)
	}
}

File: modules/opr/parse.go [hunk 5]
```diff
@@ -7,31 +7,43 @@ import (
...
 func Parse(data []byte) (OPR, error) {
-	p, err := ParseV2Content(data)
-	if err == nil {
-		return p, nil
-	}
+	// The order v1, then v2 MATTERS.
+	// When doing v2 then v1, there is valid json that decodes successfully into
+	// the protobug opr.
+	// See TestStrangeVector for the example. Since json is only valid for ascii
+	// space, it's more likely to fail given protobuf data.
```
WhoSoup: that's weird, i definitely assumed that protobuf would always fail on json input, but i can see how the "repeated" keywords might get it to accept a surplus input as valid.

modules/testutils/opr.go
```diff
@@ -0,0 +1,131 @@
...
+		winners[i] = hex.EncodeToString(b)
+	}
+	return winners
+}
+
+// FlipVersion is helpful if you want the other version than you are using
+//	1 -> 2, or 2 -> 1
+func FlipVersion(version uint8) uint8 {
+	// Invert and take the bottom 2 bits
+	return ^version & 3
```
WhoSoup: :cat: 
i would have just done `return 3-version`

Emyrk: Lol. That seems way simplier. I was thinking of the best way to do this, and just thought they were binary inverts of each other lol

modules/grader/base.go
Emyrk: True, but if we do without the `once`, then your repo that tracks the number of references will be a bit incorrect.

If we drop the `once`, then we should make a habit of always releasing it after use. Which I'm not sure is really worth it. I would feel a bit off creating multiple references to different LXR hashes.

With the sync wrapper, it's also threadsafe.

WhoSoup: pegnet wants lxrhash loaded in memory for the entire lifetime and we have no plans on ever releasing it. we're not using the module's release feature so we can create as many references as we want

modules/factoidaddress/factoidaddress.go
Emyrk: :+1:

modules/factoidaddress/factoidaddress.go
Emyrk: :+1:

modules/factoidaddress/factoidaddress.go
Emyrk: :+1:

modules/grader/base.go
Emyrk: Yes. Maybe we should change it so you can pull in the modules repo, and release it yourself. 

Like as an importer, you can `modules.Release()` or w/e. Idk if we should force the caller to call the init? But I feel weird just creating a new reference each time if we have the release feature in the lxr repo. We can't guarantee the caller will want it for it's entire lifetime for any other application.

WhoSoup: i'm not quite sure what you meant

the singletons are only inside a single application, ie how pegnet manages the singletons has no effect on any other application.

the release functionality only exists for applications that want to temporarily load lxrhash (like a wallet that only needs it once an hour or something). they're going to have to design their application around that and make sure to release their references

pegnet needs the lxr hash until the app shuts down, so we never want to release it in the pegnet app, hence we can just keep calling lxr.Init(...) over and over if we want to. every call returns the same reference

WhoSoup: update: i think i get it now. we need a way for **other applications** who are using **this module** to release **this module's instance** of lxrhash

we can either: add an explicit Release() function OR auto-release. let's discuss it in the other PR

PR #317, opened 2019-09-23
Conversions math module

sambarnes:
Add a module that contains a function that does conversion math given an amount and two exchange rates (denominated in $1 USD * 1e8)

### DIFF
--- a/modules/conversions/conversions.go
+++ b/modules/conversions/conversions.go
@@ -5,47 +5,24 @@ import (
 	"math/big"
 )
 
-// OraclePriceRecordAssetList is used such that the marshaling of the assets
-// is in the same order, and we still can use map access in the code
-// 	Key: Asset
-//	Value: Exchange rate to USD * 1e8
-type OraclePriceRecordAssetList map[string]uint64
-
-// MinimumInputNeeded tells us how many `inputType` tokens are needed to create the requested amount of
-// 'outputType` tokens.
-// All inputs must be in their lowest divisible unit as whole numbers.
-//	?? inputType -> X outputType
+// Convert takes an input amount and returns an output amount that can be created
+// from it, given the two rates `fromRate` and `toRate` denominated in 1e-8 USD.
+// All parameters must be in their lowest divisible unit as whole numbers.
+//  X fromType -> ?? toType
 //
-//  X outputType     outputRate USD     1 outputType
-//  ------------  *  --------------  *  -------------  =  ?? inputType
-//        1           1 outputType      inputRate USD
-func (o OraclePriceRecordAssetList) MinimumInputNeeded(inputType string, outputType string, requestedOutputAmount int64) (int64, error) {
-	return o.exchange(outputType, requestedOutputAmount, inputType)
-}
-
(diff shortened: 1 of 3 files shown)


modules/conversions/conversions.go
```diff
@@ -0,0 +1,47 @@
...
+	amt := big.NewInt(amount)
+
+	// Now we can run the conversion
+	// ALWAYS multiply first. If you do not adhere to the order of operations shown
+	// explicitly below, your answer will be incorrect. When doing a conversion,
+	// always multiply before you divide.
+	//  (amt * fromrate) / torate
+	num := big.NewInt(0).Mul(amt, fr)
+	num = num.Div(num, tr)
+	if max.Cmp(num) < 0 { // max < num
```
PaulBernier: Would https://golang.org/pkg/math/big/#Int.IsInt64 built-in function be better?

sambarnes: Yes, certainly would. That is a cleaner way of doing it.

modules/conversions/conversions.go
```diff
@@ -0,0 +1,44 @@
...
+	tr := big.NewInt(0).SetUint64(toRate)
+	amt := big.NewInt(amount)
+
+	// Now we can run the conversion
+	// ALWAYS multiply first. If you do not adhere to the order of operations shown
+	// explicitly below, your answer will be incorrect. When doing a conversion,
+	// always multiply before you divide.
+	//  (amt * fromrate) / torate
+	num := big.NewInt(0).Mul(amt, fr)
+	num = num.Div(num, tr)
```
PaulBernier: FYI you don't need that assignment, because `Div` modifies directly `num`

modules/conversions/conversions.go
```diff
@@ -0,0 +1,44 @@
...
+	if fromRate == 0 || toRate == 0 {
+		return 0, fmt.Errorf("invalid rate: 0")
+	}
+
+	// Convert the rates to integers. Because these rates are in USD, we will switch all our inputs to
+	// 1e-8 fixed point. The `want` should already be in this format. This should be the most amount of
+	// accuracy a miner reports. Anything beyond the 8th decimal point, we cannot account for.
+	//
+	// Uses big ints to avoid overflows.
+	fr := big.NewInt(0).SetUint64(fromRate)
```
PaulBernier: FYI if you're setting immediately a value, there is no need to call `NewInt(0)` as it makes an extra call to `.SetInt64(x)` under the hood: https://golang.org/src/math/big/int.go?s=6294:6327#L68

`new(big.Int).SetUint64(fromRate)` is enough

modules/conversions/conversions.go
Emyrk: I like this explicitness of this though.

PaulBernier: And this is exactly why immutable structures and functional programming have gained such momentum during the last decade over mutations. But I guess in golang it's aiming for performance so that makes sense, but then you should embrace the paradigm of the package and do it the mutable way. And I am pretty sure that Div returns a pointer to the structure to be able to chain operations, not to be used in an assignment.

PR #318, opened 2019-09-24
Modularized Grading

WhoSoup:
This is a pretty big PR but it's mostly independent of the develop branch. We want this to replace some of pegnet's now-redundant code but a PR like that is way too big, so the suggested process is:
1. make sure the code works (this is mostly done thanks to @Emyrk's fantastic unit test work)
2. pull the branch in
3. start a new branch that then changes the pegnet node to use the module instead of its own implementation

The key aspect is that we have an independent grader AND an independent opr module, where "independent" means that these modules DO NOT import any packages from the base pegnet node. modules only call other modules or outside repos (lxrhash counts as an outside repo) 

module/opr is responsible for:
* parsing the content of factom entries (does not have extid or height awareness)
* packaging it in a version independent interface `OPR`

module/grader is responsible for:
* grading an entire block of oprs
* verifying the content of a block (based on specified list of previous winners and version)

module/factomaddress is responsible for:
* verifying the format of an FA address

module/lxr30 is a temporary module responsible for:
* being a wrapper that allows modules and pegnet node to use the same instance of lxrhash before the lxrhash repo's singleton feature was a thing

at the moment, neither the opr nor the grader module are aware of chain-specific activation heights, meaning whoever implements these modules is responsible for initializing the grader with the correct version

Things left to do are:
* figure out the go.mod situation so that these modules can be pulled into other projects correctly. if nested go.mod files are okay
* figure out the lxr singleton release situation for the grader module


Emyrk:
Just adding a comment, a unit test was made to ensure the grading was consistent with the old grading in another branch seen here: https://github.com/pegnet/pegnet/blob/2db49ae97e4191a8c16af2fbe5b1dfec060070d7/modules/grader/grader_test.go#L23

But we are not checking that in, as the old code will be removed.

Emyrk:
I think we can drop the `module/lxr30` directory. I did update the go.mod for the latest lxr repo

modules/grader/v2_util.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if !verifyWinnerFormat(o.GetPreviousWinners(), 10) && !verifyWinnerFormat(o.GetPreviousWinners(), 25) {
+		return nil, NewValidateError("incorrect amount of previous winners")
+	}
+
+	if !verifyWinners(o.GetPreviousWinners(), winners) {
+		return nil, NewValidateError("incorrect set of previous winners")
+	}
+
```
Emyrk: We should also validate the minerid and the factoid address are valid in the v2 opr validate

PR #320, opened 2019-09-25
Set the right coinbase field

Emyrk:
We should really drop the other field. That is confusing.

### DIFF
--- a/networkMiner/client.go
+++ b/networkMiner/client.go
@@ -8,14 +8,13 @@ import (
 	"net"
 	"time"
 
-	"github.com/pegnet/pegnet/balances"
-
 	"github.com/FactomProject/factom"
+	"github.com/pegnet/pegnet/balances"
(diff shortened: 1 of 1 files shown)


networkMiner/client.go
```diff
@@ -8,14 +8,13 @@ import (
...
-	"github.com/pegnet/pegnet/balances"
-
 	"github.com/FactomProject/factom"
+	"github.com/pegnet/pegnet/balances"
 	"github.com/pegnet/pegnet/common"
 	"github.com/pegnet/pegnet/mining"
 	"github.com/pegnet/pegnet/opr"
 	log "github.com/sirupsen/logrus"
-	"github.com/zpatrick/go-config"
+	config "github.com/zpatrick/go-config"
```
Emyrk: My IDE forces me to put the `config` first, or the auto import keeps removing it.

PR #326, opened 2019-10-02
Validate minerid and factoidaddress in opr

Emyrk:


WhoSoup:
i thought it did this already, thanks for adding it!

PR #330, opened 2019-10-08
Pnt 329 import fix

Emyrk:
By moving `common.Monitor` to it's own package, and adding EC address validate to the `factoidaddress` module, we can drop the `factom` lib dependency from `pegnetd`.


Emyrk:
More work needed

PR #331, opened 2019-10-08
Pnt 328 peg source

Emyrk:


Emyrk:
> Shouldn't this PEG value check be dropped too? https://github.com/pegnet/pegnet/blob/master/opr/opr.go#L141

Currently not implemented is the activation height of PEG pricing. We need to decide an activation height, then I can drop that check at anything over the height.

I do have a fear that pegnetd could hit an issue and pause for all miners. Then all miners would be stuck, and we'd have no pegnet until we resolve pegnetd. The polling currently takes a pegnetd failure as a failed block to mine. So I think a discussion should be had around how to handle failures.

We can either have:
- Failure freeze pegnet
- Failure freeze PEG conversions, but the rest of the net continues.

Emyrk:
@sambarnes This is good for testing. I am keeping it a draft until we get an activation height and have the error conversion I described above

sambarnes:
Steven and I were talking. This feature is actually not necessary in pegnet miner for now.

First, we were talking about all the potential error cases like "What if pegnetd is unreachable or has a sync error that makes it impossible to grab newest issuances? Do we submit a 0 for PEG price? Thus halting PEG conversions until pegnetd is up and running again with a patched release?" 

Then I asked "What if a given miner submits a PEG price estimate that doesn't follow the formula (given their rates submitted and issuances from the previous block)? Would that OPR be invalid since they aren't following the rules of the pricing formula?"

Which then led to the conclusion that we shouldn't even be putting the price in the OPR because the wallets/nodes already have all the information they need to set the peg price for a given graded block. So long as the price is set by this formula, we don't need to have PEG price "estimates" and don't need to grade them.

The solution is to halt the efforts of implementing a PEG datasource for miners, and rather do the following in pegnetd's flow:
- we grade a block and find the winning rates for that block (PEG is still 0 in the OPRs)
- we calculate the PEG price given the winning rates and the asset supplies at the end of last block, and include that in the conversion rates that are committed to the database
- conversions are able to be executed for `PEG --> ?` or `? --> PEG` using said rates

All of this without needing a miner update. Granted it *is* a soft-fork / mandatory upgrade for the pegnetd node.

Thoughts @PaulBernier ?

PR #337, opened 2019-11-26
Pip 007

Emyrk:
Module that implements this: https://github.com/pegnet/pips/blob/007/pip-0007/ConversionPayout.pdf


modules/conversions/conversionlimit.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+const (
+	// PEG allocated per block for conversions (not including the bank)
+	PerBlock uint64 = 5000 * 1e8
+)
+
+// ConversionSupplySet indicates the total amount of PEG allowed to be converted
+// per block. This amount is currently set to 5,000 PEG, matching the miner
+// amount per block. (not inlcluding the bank)
```
WhoSoup: *including

modules/conversions/conversionlimit.go
```diff
@@ -0,0 +1,130 @@
...
+		// Only 1 top requester.
+		payouts[top[0]] += dust
+	} else {
+		// More than 1 with the same top amount, highest entryhash wins
+		// Sort sorts them with the lowest entryhash first
+		// If two conversions are in the same entryhash, the lowest txindex
+		// wins.
+		top = transactionid.SortTxIDS(top)
+		payouts[top[0]] += dust
+	}
```
WhoSoup: the highest transaction can be much easier calculated in AddConversion() if you store "most" and "top" in the `ConversionSupplySet`. this way is fine, though

Emyrk: Oh that is a good point :thinking:. I'm going to keep this, as I'm not worried about the cost impact of this loop. We can do that optimization down the road

PR #341, opened 2019-12-02
Free float PEG Price

Emyrk:


modules/opr/opr.go
```diff
@@ -7,7 +7,7 @@ const (
 	_ Type = iota
 	// V1 is JSON
 	V1
-	// V2 is Protobuf
+	// V2 + V3 is Protobuf
 	V2
```
Emyrk: Maybe we should rename this to `Json` and `Proto` like they were originally. Or I should copy the Proto for a v3?

cmd/cmd.go
```diff
@@ -554,6 +554,19 @@ var decodeEblock = &cobra.Command{
...
+			// fetch the eblock at the height
+			dblock, _, err := factom.GetDBlockByHeight(int64(height))
+			if err != nil {
+				CmdError(cmd, err)
+			}
+			for _, ent := range dblock.DBEntries {
+				if ent.ChainID == "a642a8674f46696cc47fdb6b65f9c87b2a19c5ea8123b3d2f0c13b6f33a9d5ef" {
+					args[0] = ent.KeyMR
+				}
+			}
```
WhoSoup: if there is no eblock found we should error-exit here, though this is an extreme edge case

cmd/root.go
```diff
@@ -62,6 +62,8 @@ func init() {
 	RootCmd.PersistentFlags().String("network", "", "The pegnet network to target. <MainNet|TestNet>")
 	RootCmd.PersistentFlags().Bool("testing", false, "Sets all activation heights to 0 so you can run on a local net")
 
+	RootCmd.PersistentFlags().Int64Var(&common.FloatingPegPriceActivation, "v3", 0, "Sets the v3 activation height")
```
WhoSoup: i think this will always override whatever we end up hardcoding. make this behave like the other flags and only override common.FPPA when set

common/common.go
```diff
@@ -29,6 +29,8 @@ func GetNetwork(network string) (string, error) {
 		return MainNetwork, nil
 	case strings.ToLower(TestNetwork), strings.ToLower("TestNet"):
 		return TestNetwork, nil
+	case strings.ToLower(UnitTestNetwork), strings.ToLower("UnitTest"):
```
WhoSoup: is there a reason to make a new network for this?

modules/grader/v3_util_test.go
```diff
@@ -0,0 +1 @@
+package grader
```
WhoSoup: could be cleaned up, doesn't have to be

modules/opr/opr.go
WhoSoup: the OPR type is different from the opr version, theoretically a json opr can hold all 3 versions and vice versa. i think it's fine to leave them as is. maybe change the comment to "V2 is for grading V2 & V3"

common/common.go
Emyrk: Yea, the unit tests were using "TestNet" before. This means the Testnet activation heights were being used, which was mainly v1 for all unit tests. I removed some of the v1 stuff from the polling, since we don't need to query XPD and XPT prices anymore. This broke the unit tests that expected the v1 asset lists. The `UnitTest` network has a helper function that takes w/e version I set it to at the start of the unit test. So now when I run unit tests, I use this network, and run a function to decide what version is "active"

cmd/root.go
Emyrk: I'm just going to drop it now. I was going to drop it later, as it was just for QA.

cmd/cmd.go
Emyrk: It will break anyway. This is mainly a debug function, and if the eblock is not found, the args[0] is set to the number, which will break on the next use of it. I can make it a cleaner exit though

WhoSoup:
approved, pending activation height

Emyrk:
I did rebase ontop of develop. there was no code changes since Who's comment.

modules/grader/v3_util.go
```diff
@@ -0,0 +1,81 @@
...
+	if len(o.Assets) != len(opr.V2Assets) {
+		return nil, NewValidateError("invalid assets")
+	}
+	for _, val := range o.Assets {
+		if val == 0 {
+			return nil, NewValidateError("assets must be greater than 0")
+		}
+	}
+
+	if len(o.Winners) != 10 && len(o.Winners) != 25 {
```
PaulBernier: you shouldn't need the 10 check as V2 already had 25 winners

PaulBernier: 10 winners should be rejected as incorrect as V2 had already 25 winners

modules/grader/v3_util.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if err := factoidaddress.Valid(o.Address); err != nil {
+		return nil, NewValidateError(fmt.Sprintf("factoidaddress is invalid : %s", err.Error()))
+	}
+
+	if valid, _ := regexp.MatchString("^[a-zA-Z0-9,]+$", o.ID); !valid {
+		return nil, NewValidateError("only alphanumeric characters and commas are allowed in the identity")
+	}
+
+	if !verifyWinnerFormat(o.GetPreviousWinners(), 10) && !verifyWinnerFormat(o.GetPreviousWinners(), 25) {
```
PaulBernier: same here about the 10 winners

modules/grader/v3_util.go
Emyrk: You are correct. Good catch

modules/grader/v3_util.go
Emyrk: You are correct. Good catch

Emyrk:
> this PR is going to require everyone to also update their configs to include the new datasources. we should use that opportunity to provide people with a "default" list that will actually make them able to earn PEGs (i think right now you need to have cmc in order to do so)
> 
> otherwise it's just some minor things/questions

I was thinking about that too, but in the past we wanted miners to make these adjustments intentionally. From what I hear, choosing data sources still matters, so there is more optimal and less optimal configurations. I'd rather not be in charge of deciding which are set.

PR #349, opened 2020-01-26
Addition of unit tests for real blocks

WhoSoup:
Since we have real V3 blocks on the Factom chain now, we can add a couple of actual blocks as part of the unit tests:

* 222270 was the activation height for V3, but it had no winners
* 222271 was the first successful V3 block
* 229145 was one of the most recent blocks when I wrote it


Emyrk:
Idk how `TestDataSourceStaleness` failed. That is the failing test. I'll look into it.

WhoSoup:
> Idk how `TestDataSourceStaleness` failed. That is the failing test. I'll look into it.

Seems like the unit test is defining "reference := time.Now()" at the beginning, but then each datasource's timestamp has a closure that calls time.Now(). the test only failed on mac and not linux, so i'd bet on a weird timer issue that only triggers if the code happens to run slow for whatever reason.

pushing a commit to see if that fixes it

Emyrk:
That fixed it. Good catch

PR #361, opened 2020-06-23
PIP - 15

StarNeit:
If one data source returns a price out of the defined tolerance band in comparison with the other sources, 
it will not allow that source’s price to be included in that block.
- Determine the Tolerance Band based on Trimmed Mean.
1. tMean >= 100000 then 0.1%
2. tMean >= 1000 then 1%
3. tMean < 1000 then 10%
- Keep the priority list and check one by one if it is existing in Tolerance Band.

### DIFF
--- a/polling/assets.go
+++ b/polling/assets.go
@@ -319,6 +319,8 @@ func (d *DataSources) PullAllPEGAssets(oprversion uint8) (pa PegAssets, err erro
 	return pa, nil
 }
 
+// Get Trimmed Mean calculation
+// https://www.investopedia.com/terms/t/trimmed_mean.asp
 func TrimmedMean(data []PegItem, p int) float64 {
 	sort.Slice(data, func(i, j int) bool {
 		return data[i].Value < data[j].Value
@@ -341,6 +343,7 @@ func TrimmedMean(data []PegItem, p int) float64 {
 //		asset		Asset to pull pricing data
 //		reference	Time reference to determine 'staleness' from
 //		sources		Map of datasources to pull the price quote from.
+//		oprversion	OPR version
 func (d *DataSources) PullBestPrice(asset string, reference time.Time, sources map[string]IDataSource, oprversion uint8) (pa PegItem, err error) {
 	if sources == nil {
 		// If our data sources passed in are nil, then we don't need to do cache wrapping.
--- a/polling/assets.go
+++ b/polling/assets.go
@@ -349,15 +349,42 @@ func (d *DataSources) PullBestPrice(asset string, reference time.Time, sources m
 		}
 	}
 
-	// Get the correct price from the idea of trimmed mean among the all enabled prices
-	// https://www.investopedia.com/terms/t/trimmed_mean.asp
-	// Medium price will eliminate the influence of outliers or data points on the tails that
-	// may unfairly affect the traditional mean.
-	if len(prices) > 0 {
-		sort.Slice(prices, func(i, j int) bool {
-			return prices[i].Value < prices[j].Value
+	pricesClone := prices
+	if len(pricesClone) > 0 {
+		// We calculate the tolerance band here, and
+		// If one datasource returns a price out of the defined tolerance band,
+		// it will not allow that source’s price to be included in that block.
+		sort.Slice(pricesClone, func(i, j int) bool {
+			return pricesClone[i].Value < pricesClone[j].Value
 		})
-		pa = prices[len(prices) / 2]
+		middle := pricesClone[len(pricesClone)/2]
+
+		toleranceRate := 0.01
+		if middle.Value >= 100000 {
+			toleranceRate = 0.001
+		}
+		if middle.Value >= 1000 {
+			toleranceRate = 0.01
+		}
+		if middle.Value < 1000 {
+			toleranceRate = 0.1
+		}
+		toleranceBandHigh := middle.Value * (1 + toleranceRate)
+		toleranceBandLow := middle.Value * (1 - toleranceRate)
+
+		// We keep datasource priority order here.
+		for i := 0; i < len(prices); i++ {
+			currentPrice := prices[i]
+			if currentPrice.Value >= toleranceBandLow && currentPrice.Value <= toleranceBandHigh {
+				return currentPrice, nil
+			}
+		}
+	}
+
+	// If we got here, that means that no price is passed by tolerance band.
+	// We will take the highest priority quote given our data-source order.
+	if len(prices) > 0 {
+		pa = prices[0]
 		return pa, nil
 	}
 
--- a/polling/assets.go
+++ b/polling/assets.go
@@ -349,8 +349,10 @@ func (d *DataSources) PullBestPrice(asset string, reference time.Time, sources m
 		}
 	}
 
-	// Get trimmed mean from all enabled prices
+	// Get the correct price from the idea of trimmed mean among the all enabled prices
 	// https://www.investopedia.com/terms/t/trimmed_mean.asp
+	// Medium price will eliminate the influence of outliers or data points on the tails that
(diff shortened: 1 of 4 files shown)


polling/assets.go
```diff
@@ -346,25 +346,16 @@ func (d *DataSources) PullBestPrice(asset string, reference time.Time, sources m
...
-	// our data sources, but they are all stale. Instead of taking the most recent price
-	// quote, we will take the highest priority quote given our data-source order.
+	// Get trimmed mean from all enabled prices
+	// https://www.investopedia.com/terms/t/trimmed_mean.asp
 	if len(prices) > 0 {
-		pa = prices[0]
+		sort.Slice(prices, func(i, j int) bool {
+			return prices[i].Value < prices[j].Value
+		})
+		pa = prices[len(prices) / 2]
```
WhoSoup, line 358: You're calculating the median, not the trimmed mean.

StarNeit: Right, but instead of using mathematical average, the median value is more reasonable.
That is because the mathematical average is not the real price which is existing on any exchanges.
Using median value sounds more perfect in several use cases, @WhoSoup

StarNeit: In this way, we will eliminate the influence of outliers or data points on the tails that may unfairly affect the traditional mean.

WhoSoup: > Right, but instead of using mathematical average, the median value is more reasonable.
> That is because the mathematical average is not the real price which is existing on any exchanges.
> Using median value sounds more perfect in several use cases, @WhoSoup

In the PR you said you were using the trimmed mean and you linked to the investopedia page on the trimmed mean both in the PR and in the comment. But you're using the median.

Also neither of this is PIP 15.

StarNeit: @WhoSoup , updated the comment.

>> Also neither of this is PIP 15.
PIP 15 is for Detecting & Preventing False Price Spikes by using defined tolerance band.
Instead of unclear tolerance band, using this solution is more reasonable for many cases.
This PR removes the False Price Spikes, and it solves the main idea of PIP 15.

WhoSoup: > PIP 15 is for Detecting & Preventing False Price Spikes by using defined tolerance band.
> Instead of unclear tolerance band, using this solution is more reasonable for many cases.
> This PR removes the False Price Spikes, and it solves the main idea of PIP 15.

PIP15 is using a tolerance band because the APIs are prioritized. The first value in the list is the one that the miner wants, the others are fallbacks. Taking the median or the mean completely negates that priority list. If the price with the highest priority is outside the tolerance, it should use the one with the next highest priority.

StarNeit: You can provide any example that doesn't work for detecting the spikes when use this solution. @WhoSoup

StarNeit: We should remove such priority list completely from the project to detect the spikes or API errors. We should give the same weights to every possible data providers, so that we could get the correct values always.

WhoSoup: The priority list exists because not all APIs are the same. There are some APIs with worse price updates (some update only once every ten minutes, for example). Those only exist as fallback and should never be picked first unless something is wrong with the best choice.

StarNeit: If miner uses wrong or bad API as their highest priority by mistake, that will make bad API result (if that is passed from the tolerance band).
I think that we must give the same weights to different data providers and get the most correct values from all possible data providers.

StarNeit: If you think that there are some APIs with "worse price" updates,
it is even not valuable to compare with them to determine the spikes.

If we give the priority to specific APIs, no one can guarantee the correct values because even that priority API can be worked wrongly.

As a solution, we should add more and more data providers and external exchanges APIs, and give them the same weights when determining the most correct values.

In that way, we can get the most correct price and remove any spikes or false API results.
@WhoSoup

WhoSoup: > You can provide any example that doesn't work for detecting the spikes when use this solution. @WhoSoup

For the fictional asset pXYZ we have three APIs: A, B, and C. A has the highest priority and provides constant updates. B has the second highest priority and updates every minute. C has the least priority and updates every 10 minutes.

Right before miners pull the data, the price goes up. You call the APIs and get the following prices:
* A: 10.5 (price from 0.01 seconds ago)
* B: 10.0 (price from 45 seconds ago)
* C: 9.99 (price from 7 minutes ago)

Both the median and trimmed mean would pick the price from B here, which is wrong. The most accurate price is from A because it provides the most up to date rate.

StarNeit: The "median or trimmed mean" don't provide the best result always.
As it is said, it is just average value, and shows the most democratic value from given inputs.
But, why it is used? It is for eliminating the influence of outliers or data points that unfairly affect the correct price.

======
Example of a Trimmed Mean
Let's say, as an example, a figure skating competition produces the following scores: 6.0, 8.1, 8.3, 9.1, and 9.9.

The mean for the scores would equal:

((6.0 + 8.1 + 8.3 + 9.1 + 9.9) / 5) = 8.28
To trim the mean by 40%, we remove the lowest 20% and the highest 20% of values, eliminating the scores of 6.0 and 9.9.

Next, we calculate the mean based on the calculation:

(8.1 + 8.3 + 9.1) / 3 = 8.50
In other words, a mean trimmed at 40% would equal 8.5 versus 8.28, which reduced the outlier bias and had the effect of increasing the reported average by 0.22 points.
=======

In worst case, "median or trimmed mean" can ignore the most correct value.
In best case, it is removing all bad spikes or false API result completely.

It is more reasonable to select latter case for safe network.
Or, we may use the trimmed mean to determine the tolerance band.

dragonheaven: If there are enough data providers or any external exchange APIs already implemented in pegnet, it sounds better if you use Trimmed Mean, @StarNeit 

@WhoSoup , how do you plan to set the tolerance band's lower/upper limit if you would like to keep the priority?

polling/assets.go
```diff
@@ -324,7 +324,7 @@ func (d *DataSources) PullAllPEGAssets(oprversion uint8) (pa PegAssets, err erro
 //		asset		Asset to pull pricing data
```
sigrlami, line 324: Can you add a description for the function, so it shows on GoDoc properly

polling/assets.go
```diff
@@ -324,7 +324,7 @@ func (d *DataSources) PullAllPEGAssets(oprversion uint8) (pa PegAssets, err erro
...
 //		reference	Time reference to determine 'staleness' from
 //		sources		Map of datasources to pull the price quote from.
-func (d *DataSources) PullBestPrice(asset string, reference time.Time, sources map[string]IDataSource) (pa PegItem, err error) {
+func (d *DataSources) PullBestPrice(asset string, reference time.Time, sources map[string]IDataSource, oprversion uint8) (pa PegItem, err error) {
```
sigrlami, line 327: Can you add a description for the function, so it shows on GoDoc properly?

polling/assets.go
```diff
@@ -346,23 +346,45 @@ func (d *DataSources) PullBestPrice(asset string, reference time.Time, sources m
...
+		}
+	}
 
-			// We can break out if this is a 'good' price
-			//	Is it stale AND the market is open? We can expect stale quotes in a closed market
-			if reference.Sub(price.When) > d.staleDuration && IsMarketOpen(asset, reference) {
-				// This price quote is stale, keep fetching prices
-				continue
+	if oprversion == 5 {
+		pricesClone := prices
```
sigrlami, line 353: why cloned version?

StarNeit: We should keep the original priority in the prices list.

dragonheaven:
Looks good to me. @StarNeit

StarNeit:
Merged. @dragonheaven

PR #365, opened 2020-07-15
PIP13 - Staking (SPR Chain) 

dragonheaven:
The goal of this PR is to counterbalance the submission of Oracle Price Records of miners by allowing stakeholders to submit prices as well. 
Prices are pulled at minute one, same time as the miners. 
SPRs are written to a new chain (PegNet / MainNet / StakingPriceRecords) at minute 8.


spr/spr.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// CreateSPREntry will create the entry from the EXISITING data.
+// It will not set any fields like in `GetSPRecord`
+func (spr *StakingPriceRecord) CreateSPREntry() (*factom.Entry, error) {
+	var err error
+	e := new(factom.Entry)
+	e.ChainID = hex.EncodeToString(base58.Decode(spr.SPRChainID))
+
+	//  Todo: verify every options
```
sigrlami, line 258: Is there an update to this? Where is option verification code?

staking/entryWriter.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (w *EntryWriter) ECBalance() (int64, error) {
+	return factom.GetECBalance(w.ec.String())
+}
+
+// PopulateECAddress only needs to be called once
+func (w *EntryWriter) PopulateECAddress() error {
+	// Get the Entry Credit Address that we need to write our SPR records.
+	if ecadrStr, err := w.config.String("Miner.ECAddress"); err != nil {
```
sigrlami, line 54: here error indicates the same behavior for different logic, for non-populated address and address missing in the config, it should be distinguished and maybe forward the explicit error to the user.

staking/entryWriter.go
```diff
@@ -0,0 +1,196 @@
...
+	w.Do(func() {
+		if blocking {
+			w.collectAndWrite()
+		} else {
+			go w.collectAndWrite()
+		}
+	})
+}
+
+// collectAndWrite is idempotent
```
sigrlami, line 98: It has side-effect by writing to blockchain, idempotent incorrect wording here. Please add more broad comment on what it does

staking/entryWriter.go
```diff
@@ -0,0 +1,196 @@
...
+func (w *EntryWriter) collectAndWrite() {
+	err := w.EntryWritingFunction() // Write to blockchain
+	if err != nil {
+		log.WithError(err).Error("Failed to write staking record")
```
sigrlami, line 102: should there be some kind of retry or identification of the error? Like not enough EC or network connectivity issue.

staking/sprmaker.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/zpatrick/go-config"
+)
+
+type ISPRMaker interface {
+	NewSPR(ctx context.Context, dbht int32, config *config.Config) (*spr.StakingPriceRecord, error)
+}
+
+// SPRMaker
+// TODO: Should we change this at all?
+type SPRMaker struct {
```
sigrlami, line 15: why empty struct?

staking/entryWriter.go
dragonheaven: panic() is just like an exception, it also arises at runtime. Regarding the error, it is processed by panic(err).
https://github.com/dragonheaven/pegnet/blob/feature/pip13/staking/coordinator.go#L39

If you just are going to suggest that we should replace all panic() by alert message to users, it should be done in another pull request.

Exception has been processed like all other errors int this project.
panic() is just enough for this moment.

spr/spr.go
dragonheaven: Do you see this PR is Draft? It is for verify options for SPREntry, @sigrlami

staking/entryWriter.go
dragonheaven: What is incorrect wording here? This collectAndWrite is for writing CreateSPREntry.

staking/entryWriter.go
dragonheaven: Retry() is not allowed with any kinds of errors with Staking modules.
In PIP13, we try to write the SPR at minute 1, and write them to SPR chain at minute 8.
All is processed based on synchronization within 10 minutes.

If there is any errors such as insufficient EC, that should be terminated until staker fixes his configuration.

staking/sprmaker.go
dragonheaven: Empty struct struct{} is realized in special way in Go.

1) It’s a smallest building block in Go. It’s size is literally 0 bytes.
2) If has zero size. you may create a slice of 1000’s empty structures and this slice will be very tiny. Because really Go stores only a number of them in the slice but not them itself. The same story with channels.
3) All pointers to it always point to the same special place in memory.
4) Very useful In channels when you have notify about some event but you don’t need to pass any information about it, only a fact. Best solution is to pass an empty structure because it will only increment a counter in the channel but not assign memory, copy elements end so on. Sometime people use Boolean values for this purpose, but it’s much worse.
5) Zero size container for methods. You may want have a mock for testing interfaces. often you don’t need data on it just methods with predefined input and output.
6) Go has no Set object. Bit can be easily realized as a map[keyType]struct{}. This way map keeps only keys and no values.

dragonheaven: In this code, empty structs behave just like any other type, it works as method receivers. @sigrlami

spr/spr.go
sigrlami: I have new interface and didn't noticed it's a draft. Thanks for pointing out.

staking/entryWriter.go
sigrlami: Yes, now it makes sense. Thanks for thorough explanation.

PR #370, opened 2020-08-17
Feature/spr grade

StarNeit:
- New SPR Grade algorithm
- Update SafeMarshal with latest SPR version
- S1Payout for top 100 staker's price rewards
- Update price record version validation
- bug fixes..

### DIFF
--- a/common/activation.go
+++ b/common/activation.go
@@ -42,10 +42,10 @@ var (
 	// StakingHeights indicates the SPR version, which dictates the SPR format.
 	StakingHeights = map[string]func(height int64) uint8{
 		MainNetwork: func(height int64) uint8 {
-			return 0 // Latest code version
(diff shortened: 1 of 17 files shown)


dragonheaven:
looks good to me. @StarNeit

common/activation.go
```diff
@@ -42,10 +42,10 @@ var (
 	// StakingHeights indicates the SPR version, which dictates the SPR format.
 	StakingHeights = map[string]func(height int64) uint8{
 		MainNetwork: func(height int64) uint8 {
-			return 0 // Latest code version
+			return 5 // Latest code version
```
dragonheaven, line 45: why staking version is 5 initially? do you wanna to keep the same with OPR?

dragonheaven: Yes, that is true.

PR #376, opened 2020-08-21
Check Addresses early, report EC balance at boot

PaulSnow:
Checks addresses for staker early, so errors are caught quickly.


PaulSnow:
I have no idea how this change fails to build.

PR #386, opened 2020-08-25
Support signing SPRs and validating signatures.

aktary:
1) Minor: needed to update the version of factom we were depending on.  As a result, there's a minor change to `factom.GetTransaction()` in burns.go.
2) Since we have existing SPRs we have to honor, we need to set an activation height for signing, thus the new Version and update to activation.go.
3) Used `factom.SignData()` to request that the wallet sign the SPR... this _should_ work, but is untested so far.  We do this signging request in `CreateSprEntry()` to support multiple addresses.
4) Validate signature in `ValidateS1()`

### DIFF
--- a/balances/burns.go
+++ b/balances/burns.go
@@ -48,7 +48,7 @@ func (b *BurnTracking) UpdateBurns(c *config.Config, startBlock int64) error {
 		}
 
 		for _, txid := range fc.Transactions {
-			txInterface, err := factom.GetTransaction(txid.TxID)
+			txInterface, err := factom.GetTransaction(string(txid))
 			if err != nil {
 				return err
 			}
--- a/balances/burns.go
+++ b/balances/burns.go
@@ -48,7 +48,7 @@ func (b *BurnTracking) UpdateBurns(c *config.Config, startBlock int64) error {
 		}
 
 		for _, txid := range fc.Transactions {
-			txInterface, err := factom.GetTransaction(string(txid))
+			txInterface, err := factom.GetTransaction(txid.TxID)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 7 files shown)


spr/spr.go
```diff
@@ -236,11 +237,21 @@ func (spr *StakingPriceRecord) CreateSPREntry() (*factom.Entry, error) {
...
-	e.ExtIDs = [][]byte{{spr.Version}, rcd, spr.GetHash()}
+
 	e.Content, err = spr.SafeMarshal()
 	if err != nil {
 		return nil, err
 	}
+
+	signature, err := factom.SignData(spr.CoinbaseAddress, e.Content)
+	if err != nil {
+		return nil, err
```
dragonheaven, line 248: I see ` Method not found` error in this code line. @aktary

aktary: Resolved by using the latest alpine factom-walletd docker image.  docker-compose.yml updated.

balances/burns.go
```diff
@@ -48,7 +48,7 @@ func (b *BurnTracking) UpdateBurns(c *config.Config, startBlock int64) error {
 		}
 
 		for _, txid := range fc.Transactions {
-			txInterface, err := factom.GetTransaction(txid.TxID)
+			txInterface, err := factom.GetTransaction(string(txid))
```
dragonheaven, line 51: Is there any special reason you changed the txid format here?

modules/graderStake/s1_util.go
```diff
@@ -40,6 +42,13 @@ func ValidateS1(entryhash []byte, extids [][]byte, height int32, content []byte)
 	}
 	o := &spr.S1Content{V2Content: *o2}
 
+	if exids[0][0] == 6 {
```
dragonheaven, line 45: variable is not correct. it should be `extids ` instead of `exids`.  @aktary

aktary: 😵

StarNeit:
This has been merged into master already, thanks

PR #389, opened 2020-08-26
SPR signature

StarNeit:
- SPR signature
- S2 Validation
- New activation height
- ComputeRCDFromPubkey


modules/graderStake/s1_util.go
```diff
@@ -40,6 +42,14 @@ func ValidateS1(entryhash []byte, extids [][]byte, height int32, content []byte)
 	}
 	o := &spr.S1Content{V2Content: *o2}
 
+	if extids[0][0] == 6 {
```
aktary, line 45: Is this still relevant if we're splitting out version 6 signature checking into s2_util?

StarNeit: `ValidateS1()` is updated.

PR #394, opened 2020-09-04
[Feature] Update datasource priority order modules

dragonheaven:
Replace for/range into for-loop index based


WhoSoup (commented):
This only works if the source with the highest priority has a lower value than the other one. You're sorting the slice before checking it so index 0 isn't always the one with the higher priority.

dragonheaven:
> This only works if the source with the highest priority has a lower value than the other one. You're sorting the slice before checking it so index 0 isn't always the one with the higher priority.

Updated. Not it checks the condition before the sorting.

PR #398, opened 2020-10-14
Test Net environment

dragonheaven:


StarNeit:
We already built testnet environment. It is not necessary.

PR #408, opened 2021-02-10
Fix randomization

dragonheaven:
- Remove randomization from 2.0.2

### DIFF
--- a/modules/graderStake/baseGradedBlock.go
+++ b/modules/graderStake/baseGradedBlock.go
@@ -1,10 +1,5 @@
 package graderStake
 
-import (
-	"math/rand"
-	"time"
-)
-
 // baseGradedBlock is an spr set that has been graded
 type baseGradedBlock struct {
 	sprs   []*GradingSPR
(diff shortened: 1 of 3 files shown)


modules/graderStake/baseGradedBlock.go
```diff
@@ -64,17 +59,4 @@ func (b *baseGradedBlock) filterDuplicates() {
 	b.sprs = filtered
 }
 
-// randomize the SPRs order.
-func (b *baseGradedBlock) shuffleSPRs() {
```
PaulSnow, line 68: Should add comments.  Suggested:

// randomize the SPRs order.  This provides the reward to a random qualified submission
func (b *baseGradedBlock) shuffleSPRs() {
	// First of all, let's make sure all instances of pegnetd use the same order for the sprs
	// We sort them by their entry hashes
	sort.Slice(b.sprs, func(i, j int) bool {
		return bytes.Compare(b.sprs[i].EntryHash, b.sprs[j].EntryHash) < 0
	})

	// Now we want to create a seed from the first 24 bits of all the accepted SPRs.
	// We will combine them in a way that all the bits in each 24 bits of an Entryhash contribute
	// the the randomness of our selection.
	var seed int64
	for _, spr := range b.sprs {
		// Mix up the bits in seed by replacing it with its bits scrambled.
		seed = seed>>1 ^ seed<<5
		// Now take the top 24 bits of spr's entryhash and xor it in.
		seed ^= int64(spr.EntryHash[0]) ^ int64(spr.EntryHash[1])<<8 ^ int64(spr.EntryHash[2])<<16
		// Each time through the loop, bits of each 24 bits of each entryhash are scrambled again and again.
		// This is a pretty solid sort of pseudorandom number generation. see https://en.wikipedia.org/wiki/Xorshift
	}
	
	// Now we randomize b.sprs by using our random number generator and telling our sort that 
	// spr[i] < spr[j] not by comparing them, but looking at the result of a stream of random numbers
	sort.Slice(b.sprs, func(i, j int) bool {
		seed = seed>>7 ^ seed<<3
		return seed&0xFF > 0x7F
	})
}

PR #409, opened 2021-02-25
Version 2.0.3

StarNeit:
- Removed randomization module of SPRs from consensus algorithm.
- Restored s2grader version to old one.


dragonheaven (approved):
Approved.

PR #411, opened 2021-02-26
Pnt 410 better feedback while staking

PaulSnow:
Provides some better error reporting and feedback.  Prints warnings rather than panics in a couple cases.


PaulSnow:
Jumps the gun.  Need to get the other pull requests in first.

PR #414, opened 2021-04-30
PIP-18:  Delegated Proof of Stake

dragonheaven:
1. PIP-18 use balances of PEG for each address which is more complicated.  It is the balance of PEG for the address (assuming it has not be delegated) 
2. It quits looking at the rich list, and just consider the top 100 submissions with the highest stake
3. PIP-18 pays out with the ratio of the total PEG staked. (Removes old top 100 PEG addresses staking reward and give staking opportunity to all PEG holders)


modules/graderStake/s4GradedBlock.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// WinnerAmount returns the version specific amount of winners.
+func (g *S4GradedBlock) WinnerAmount() int {
+	return 100
+}
+
+// Winners returns the winning SPRs
+func (g *S4GradedBlock) Winners() []*GradingSPR {
+	if len(g.sprs) < 100 {
```
PaulSnow: switch cnt := len(g.sprs) {
   case cnt < 25:
        return nil
   case cnt > 100:
        return g.sprs[:100]
   default:
        return g.sprs
}

PaulSnow: I don't think we have to have 100 addresses.  I believe 25 is what we require?

modules/graderStake/s4GradedBlock.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil
+	}
+
+	return g.sprs[:100]
+}
+
+func (g *S4GradedBlock) grade() {
+	if len(g.sprs) < 100 {
```
PaulSnow: I don't think we want to require 100 sprs.  I believe now the requirement is 25

modules/graderStake/s4grader.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	s4.sprs = append(s4.sprs, gspr)
+	return nil
+}
+
+// Grade the SPRs. The S1 algorithm works the following way:
+//	1. Take the top 50 entries with the best proof of work
+//	2. Remove top and low's 1% band from each of the 32 assets
+//	3. Calculate the average of each of the 32 assets
+//	4. Calculate the distance of each SPR to the average, where distance is the sum of quadratic differences
```
presidento23, line 36: I don't see steps 2-6 taking place. if you mind breaking down what you did that would be much appreciated

dragonheaven: That is grading algorithm which were implemented from version 1.0.
No need to discuss about that grading algorithm in this PR.

